Issue #4, opened 2016-10-24
Add homogeneous intersection functions


ianmackenzie:
- [ ] `Axis2d.intersection : Axis2d -> Axis2d -> Maybe Point2d`
- [ ] `Plane3d.intersection : Plane3d -> Plane3d -> Maybe Axis3d`
- [ ] `LineSegment2d.intersection : LineSegment2d -> LineSegment2d -> Maybe Point2d`
- [ ] `Triangle3d.intersection : Triangle3d -> Triangle3d -> Maybe LineSegment3d`

mpizenberg:
Hi, I mentionned last time in elm discuss that I had some addition to LineSegment2d and it was for the intersection so glad to see that it was already planned :). The interface I implemented though is a bit different. What I did was with this interface:
```elm
type RelativePosition
    = Parallel
    | Collinear
    | Intersection Point2d
    | NonIntersection Point2d

relativeTo : LineSegment2d -> LineSegment2d -> RelativePosition
```
It includes your interface by saying `Intersection Point2d` is equivalent to `Just Point2d` and the rest is equivalent to `Nothing`. So if you don't like the additional stuff, I can definitely simplify my implementation. Or keep it private and implement the `intersection` function on top of it.

ianmackenzie:
Does your current implementation use any kind of tolerance, or (for example) are line segments only considered to be parallel if they have *exactly* the same direction? I suspect in many cases if you care about checking for parallel or collinear line segments you will want to do so within a tolerance, but exactly how that tolerance should be defined is not clear to me (especially for collinearity). How was your client code (the code using your interface) using the various cases of the union type?

Oh, and it couldn't be called `relativeTo`, in OpenSolid that name is taken =) Perhaps something like `classify`?

mpizenberg:
Good point, currently only exact parallelism is taken into account (same for collinearity). I guess another function with an angle tolerance would make sense, but also make the computation a little heavier since cross product would not be enough. A normalization would be required to have reliable angle thresholds.

I confess, for now my only use of `relativeTo` was in the polygon self intersection algorithm. And it only searched for intersection:
```elm
eventIntersection : Event -> Event -> Maybe Point2d
eventIntersection ( id1, _, seg1 ) ( id2, _, seg2 ) =
    if cyclicDistance id1 id2 <= 1 then
        Nothing
    else
        case LineSegment2d.relativeTo seg1 seg2 of
            LineSegment2d.Intersection point ->
                Just point

            _ ->
[...]

PR #17, opened 2017-03-06
Homogeneous LineSegment2d intersection (#4).

mpizenberg:
Possible implementation of the homogeneous intersection function for `LineSegment2d`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,4 +40,4 @@ before_script:
 
 script:
   - cd $TRAVIS_BUILD_DIR/tests
-  - elm-test --seed 948576658 Main.elm
+  - elm-test Main.elm
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,4 +40,4 @@ before_script:
 
 script:
   - cd $TRAVIS_BUILD_DIR/tests
-  - elm-test Main.elm
+  - elm-test --seed 948576658 Main.elm
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,3 +3,4 @@ the opensolid/geometry package. If you submit a patch to this repository, please
 add your name to the end of this list!
 
 Ian Mackenzie <ian.e.mackenzie@gmail.com>
+Matthieu Pizenberg <matthieu.pizenberg@gmail.com> (intersections)
--- a/src/OpenSolid/LineSegment2d.elm
+++ b/src/OpenSolid/LineSegment2d.elm
@@ -286,9 +286,11 @@ intersection lineSegment1 lineSegment2 =
     -- p |--- r ---| p+r
     -- q |--- s ---| q+s
     let
-        ( p, q ) =
+        ( p, q, p_, q_ ) =
             ( startPoint lineSegment1
             , startPoint lineSegment2
+            , endPoint lineSegment1
+            , endPoint lineSegment2
             )
 
         ( r, s, pq ) =
@@ -306,6 +308,10 @@ intersection lineSegment1 lineSegment2 =
         if rXs == 0 then
             -- parallel or collinear lines
             Nothing
+        else if p == q || p == q_ then
+            Just p
+        else if p_ == q || p_ == q_ then
+            Just p_
         else
             let
                 ( t, u ) =
--- a/src/OpenSolid/LineSegment2d.elm
+++ b/src/OpenSolid/LineSegment2d.elm
@@ -259,22 +259,25 @@ is no such point (the two line segments do not touch, or they overlap), returns
 `Nothing`.
 
     -- 4 corners of a square
-    a = Point2d (0, 0)
-    b = Point2d (1, 0)
-    c = Point2d (1, 1)
-    d = Point2d (0, 1)
+    ( a, b, c, d ) =
+        ( Point2d (0, 0)
+        , Point2d (1, 0)
+        , Point2d (1, 1)
+        , Point2d (0, 1)
+        )
 
     -- definition of some segments with those points
-    ab = LineSegment2d (a,b)
+    ab =
+        LineSegment2d ( a, b )
     ...
 
     -- searching for intersections
 
     LineSegment2d.intersection ab bc
-    --> Just (Point2d (1,0)) -- corner point b
+    --> Just (Point2d ( 1, 0 )) -- corner point b
 
     LineSegment2d.intersection ac bd
-    --> Just (Point2d (0.5,0.5)) -- diagonal crossing at square center
+    --> Just (Point2d ( 0.5, 0.5 )) -- diagonal crossing at square center
 
     LineSegment2d.intersection ab cd
     --> Nothing -- parallel lines
--- a/src/OpenSolid/LineSegment2d.elm
+++ b/src/OpenSolid/LineSegment2d.elm
@@ -341,9 +341,6 @@ intersection lineSegment1 lineSegment2 =
                     , pqXr / uDenominator
                     )
 
-                intersection =
-                    interpolate lineSegment1 t
-
                 within start end x =
                     start <= x && x <= end
             in
--- a/src/OpenSolid/LineSegment2d.elm
+++ b/src/OpenSolid/LineSegment2d.elm
@@ -347,6 +347,9 @@ intersection lineSegment1 lineSegment2 =
                 if within 0 1 t && within 0 1 u then
                     -- Intersection is within both segments.
                     let
+                        -- Ensure interpolation happens from the closest
+                        -- endpoint (this should be more numerically stable, and
+                        -- also mostly ensures that intersection is symmetric)
                         intersection =
                             if min t (1 - t) <= min u (1 - u) then
                                 interpolate lineSegment1 t
--- a/src/OpenSolid/LineSegment2d.elm
+++ b/src/OpenSolid/LineSegment2d.elm
@@ -286,12 +286,11 @@ intersection lineSegment1 lineSegment2 =
     -- p |--- r ---| p_
     -- q |--- s ---| q_
     let
-        ( p, q, p_, q_ ) =
-            ( startPoint lineSegment1
-            , startPoint lineSegment2
-            , endPoint lineSegment1
-            , endPoint lineSegment2
-            )
+        ( p, p_ ) =
+            endpoints lineSegment1
+
+        ( q, q_ ) =
+            endpoints lineSegment2
 
         ( r, s, pq ) =
             ( vector lineSegment1
--- a/src/OpenSolid/LineSegment2d.elm
+++ b/src/OpenSolid/LineSegment2d.elm
@@ -19,6 +19,7 @@ module OpenSolid.LineSegment2d
         , reverse
         , midpoint
         , interpolate
+        , intersection
         , length
         , squaredLength
         , direction
@@ -66,6 +67,10 @@ the `LineSegment2d` constructor, for example
 # Interpolation
 
 @docs midpoint, interpolate
+
+# Intersection
+
+@docs intersection
 
 # Length and direction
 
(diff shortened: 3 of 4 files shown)


ianmackenzie:
Just made some modifications to your PR branch, take a look...one wrinkle I'd ideally like to resolve is that right now calling `LineSegment2d.intersection` on two line segments which share an endpoint does not have consistent behaviour. Sometimes the shared endpoint is reported as the intersection point and sometimes not. This isn't surprising, due to numerical roundoff, but I'd like to tweak the logic so that there *is* consistent behaviour if possible.

I've added a couple tests, one that checks that shared endpoints are *always* found and one that checks that they are *never* found - obviously we can only keep one of them, and I think I'm actually leaning towards "never find shared endpoints" to emphasize that this is really just for finding true *crossing* intersections.

Thoughts on how this might be done? One way I can think of is to check which pair of start/end points of the two segments are closest to each other and use that pair as the two start points (basically, reverse one or both line segments before proceeding with the algorithm to ensure that the two start points are as close together as possible). But that would add some measurable overhead...

ianmackenzie:
Also, I should have started with this, but: thank you! First pull request, very exciting =)

I've just added an AUTHORS file in the master branch - please feel free to add yourself to it as part of this PR.

mpizenberg:
Thank you for the "AUTHORS" offer ;) it's a rather small contribution but hey, better small than nothing ^^.
Regarding the code modifications I indicated in the commits comments, I would like first to make the tests execute on my computer. I have never used elm test before. A little help with that, maybe as a section in `CONTRIBUTION.md` would be welcomed.

ianmackenzie:
> Maybe the whole "Interpolation" block should also move after length and direction?

I'm not sure there's any one obvious answer for how to group things (to me `interpolate` and `midpoint` belong together, and `midpoint` belongs near `startPoint`  and `endPoint`...). Maybe after adding some more functions there will be some more obvious patterns to use, but let's just leave it as it is for now.

> If I name t and u the position weights on each respective segment, I just have to check that 0+epsilon < t < 1 - epsilon and same for u.

I'd really prefer to avoid any hardcoded tolerance values in the code, since in my experience they're very difficult to get right and more often than not lead to bugs several months later when you encounter some particularly weird (but valid) geometry. I'm currently tempted to just add an initial four equality checks between the pairs of endpoints of the two line segments - if any two endpoints are equal, then there can be no crossing intersection.

> I would like first to make the tests execute on my computer. I have never used elm test before. A little help with that, maybe as a section in CONTRIBUTION.md would be welcomed.

Excellent point! I've updated CONTRIBUTING.md, and added support for running the tests at the console with `elm-test` while I was at it (I've just been using the HTML test runner, which doesn't require installing any NPM packages, but now you have both options). Next step is setting up a Travis build...

mpizenberg:
> I'm currently tempted to just add an initial four equality checks between the pairs of endpoints of the two line segments - if any two endpoints are equal, then there can be no crossing intersection.

After some interior debate, it makes more sense to me that a shared point is considered an intersection point. When thinking of segments, we refer to the "closed" bracket version `[ab]` so an end point belong to the two segments, and by definition (except in the parallel case) is the intersection point. Did you have in mind some application where not finding end points makes more sense?

> Excellent point! I've updated CONTRIBUTING.md

Thanks, it works, that's awesome!

> Next step is setting up a Travis build...

According to [elm-test on npm](https://www.npmjs.com/package/elm-test#travis-ci), the github file side is easy (just one yaml file). I've never set up a Travis build though. Might be worth trying when I have time if you've not done it yet.

ianmackenzie:
> After some interior debate, it makes more sense to me that a shared point is considered an intersection point. When thinking of segments, we refer to the "closed" bracket version [ab] so an end point belong to the two segments, and by definition (except in the parallel case) is the intersection point. Did you have in mind some application where not finding end points makes more sense?

It's true that if we think of `intersection` as a mathematical set intersection between the sets of points on the two line segments, then coincident endpoints should be included. But I think defining `intersection` as "the point (if it exists) where the two segments cross over each other" is a better and more useful definition, for a few reasons:

  - It naturally excludes the overlap case, since overlapping segments do not cross each other. Defining `intersection` as a pure mathematical set intersection implies that the intersection of two overlapping segments should return the overlapping segment, but that both complicates the API (would need a custom return type like the one you defined) and is numerically unreliable (non-axis-aligned line segments may or may not appear to be _exactly_ collinear because of floating-point roundoff). The alternative would be to define `intersection` as "the point common to the two segments, if it exists and is unique", which is not awful but I still think is not great, for the next two reasons.
  - Using "cross over" in the definition makes it clear that `intersection` should not be used to check for points where line segments just touch each other without crossing. If we say that `intersection` will find touching endpoints, then it makes sense to think that perhaps it will also find points where two line segments just touch each other in a 'T' configuration. However, there's no real good way to check for that case without using some sort of tolerance (except in special cases like axis-aligned segments). If someone does want to check for that case specifically, then I think it makes more sense to do it with a tolerance using some separate function (like checking endpoints of one segment against the other segment using a `Point2d.isOnLineSegmentWithin : Float -> LineSegment2d -> Point2d -> Bool` or similar function).
  - It seems to me that there are several cases where you only want to find crossing intersections (including checking for self-intersections in polygons/polylines!), and fewer where you want a general `intersection` function to also find special cases like coincident endpoints. If you do want to check for coincident endpoints, it's easy enough to add a separate check for that (which you could also customize separately by doing things like checking endpoints for equality within a tolerance) - seems more annoying to get all intersections and then have to check for and filter out coincident-endpoint intersections if you don't want them.

Thanks for your perseverance keeping up with all this discussion! It seems like a lot of work over one fairly small function, but I'd really like to get it right - and if we can figure out some good general design principles for `LineSegment2d.intersection`, then it should make designing and implementing future intersection functions easier.

ianmackenzie:
> According to elm-test on npm, the github file side is easy (just one yaml file). I've never set up a Travis build though. Might be worth trying when I have time if you've not done it yet.

Travis build is now set up, and I've merged the relevant commits from master into the pull request branch so everything should now work.

mpizenberg:
> The alternative would be to define intersection as "the point common to the two segments, if it exists and is unique"

I'd say something like: "the point common to two non parallel segments, if it exists".

> I think defining intersection as "the point (if it exists) where the two segments cross over each other" is a better and more useful definition, for a few reasons.
...
Using "cross over" in the definition makes it clear that intersection should not be used to check for points where line segments just touch each other without crossing.

Perhaps functions named `intersection` and `crossing` could coexist with slightly different specifications.
```elm
-- The point (if it exists) common to two non parallel segments.
intersection : LineSegment2d -> LineSegment2d -> Maybe Point2d

-- The point (if it exists) where the two segments cross over each other.
-- A crossing distance threshold must be specified
crossing : Float -> LineSegment2d -> LineSegment2d -> Maybe Point2d
```

> It seems to me that there are several cases where you only want to find crossing intersections (including checking for self-intersections in polygons/polylines!), and fewer where you want a general intersection function to also find special cases like coincident endpoints.

Well, I would say the opposit ^^. It seems to me that there are several cases where algorithms rely on detecting those intersections. In the polygon self intersection algorithm, for exemple, you want to detect an `X` configuration (4 segments, a point in the center of the X) or even an `><` configuration where segments touch each other and then go back. If you deterministically don't detect the `><` or `X` for exemple, you will not know that your polygon is not single, and then there are plenty of other algorithms that you can't perform reliably because they only work well on simple polygons (like some filling algorithms, etc.). In case the segment intersection is deterministically not detecting sharing points, then you would have to detect them another way. Then comes another issue the collinear case. If two segments are colinear and share one point, in the additionnal step, I would also have to check that segments are not parallel before retaining the sharing point. It feels like I'm computing everything twice the number that should be needed.
PS, on wikipedia in the dealing with [numerical issues paragraph of the Bentley-Ottmann algorithm](https://en.wikipedia.org/wiki/Bentley%E2%80%93Ottmann_algorithm#Special_position_and_numerical_precision_issues), it is mentionned that detecting end points is a good idea.
> Define a line segment to be a closed set, containing its endpoints. ...

The drawback is that we don't have a deterministic behavior with `T` configuration (which by the way is also the case if we just ignore sharing points instead of keeping them). However this means that the middle point would be considered either on, over, or below the top T bar but not all at the same time. So in the `X` case mentionned before or an `-|-` case, there is no issue since at least one pair of segments intersection should be detected.

> Thanks for your perseverance keeping up with all this discussion! It seems like a lot of work over one fairly small function, but I'd really like to get it right - and if we can figure out some good general design principles for LineSegment2d.intersection, then it should make designing and implementing future intersection functions easier.

Yep, totally agree with this. By the way, I have Breton origins and we are told to be a bit too stuborn ^^. But please don't misinterpret my perseverance with childish stubbornness. I'm here truly because your work is awesome and I also wish to make the best design decisions.

ianmackenzie:
I tried a tweak to the algorithm logic that avoids needing the special-case equality check by ensuring that parameter values of 0 or 1 will always be computed exactly - see what you think. Although if we want to go with 'always find coincident endpoints' then I think there needs to be some extra special-case logic (regardless of whether we use my modified version or not) in the 'parallel/collinear' case since that may actually result from one or both line segments being zero length. In that case, presumably we would just fall back to checking endpoints.

> Perhaps functions named intersection and crossing could coexist with slightly different specifications.

I'm not sure one needs a tolerance and one doesn't - if we go with two, I'm thinking maybe `intersection` and `strictIntersection` with the same signature, but the latter of which doesn't include endpoints (i.e. the intersection point must be strictly inside each segment).

> So in the X case mentionned before or an -|- case, there is no issue since at least one pair of segments intersection should be detected.

I'd like to add a special test to check for this, particularly the -|- case - I'd like to make sure there are no weird cases where the segments on both sides both 'round out' somehow and both don't get seen as intersecting. Ideally, I think the two segments on each side should either:

  - Both find the same intersection point, which is equal to their shared endpoint.
  - Have only one find an intersection point, which should *not* be equal to the shared endpoint (that is, numerical roundoff makes the shared endpoint appear to be slightly to one side of the line, but this is seen consistently by both segments).

Unfortunately I'll be mostly offline for the next few days - I'm refereeing at [a robotics competition](http://www.firstinspires.org/robotics/frc) this weekend. I'd like to think a bit more about the rest of your comments, but it might be a few days before I get a chance to reply.

mpizenberg:
> I'd like to add a special test to check for this, particularly the -|- case

Agreed 👍 

> Unfortunately I'll be mostly offline for the next few days - I'm refereeing at a robotics competition this weekend. I'd like to think a bit more about the rest of your comments, but it might be a few days before I get a chance to reply.

Cool :) that will be fun! I used to be on the student side of this. (Well not the "first" competition but we also have in France an [annual student robotics competition](http://www.eurobot.org/eurobot/eurobot-2017)). Enjoy!

ianmackenzie:
OK, I think you've convinced me that endpoints should be included...I've added some tests for some collinear-segment cases and some extra special-case logic to handle those cases. See what you think! I think this is getting close...although I'd still like to add some tests to make sure intersection behaves in a consistent way in the -|- case where two line segments have a shared endpoint which lies exactly (or almost exactly) on a third segment, and I might tweak the documentation a bit to better match the style of the rest of the documentation.

mpizenberg:
> I'd still like to add some tests to make sure intersection behaves in a consistent way in the -|- case where two line segments have a shared endpoint which lies exactly (or almost exactly) on a third segment

Done. You might want to check that I didn't make a mistake though.

Also, while doing some tests, I realized that the test `intersectionOfEqualLineSegmentsIsNothing` was sometimes failing. It had to do with cases where the segment was degenerated (reduced to a point). In that case, the recently added new logic of a shared end point was detecting the point. So I tweaked this test a bit in b96aad8 and added a dedicated test for the degenerated case in 2494108. See what you think about that.

Since we are almost reaching the merging point, I added my name in the Authors list. Thanks again for your patience and guidance.

ianmackenzie:
That all looks great! I've made a few more tweaks - check them out and see what you think. Unless you have any concerns about those modifications or have anything else you'd like to change, I think this is ready to merge!

mpizenberg:
Looks good to me. I corrected the ReversedEqualLineSegment case that I forgot when correcting the Equal test last time. Weirdly this reversed test never used degenerate segments with fuzz tests, and so never failed previously apparently. Also I made the -|- test case even stricter by specifying what to expect in case both points are on the same side of the third segment (the >| case) instead of just pass.

ianmackenzie:
Looks great! One more question I thought of just before merging - what are your thoughts on renaming `intersection` to `intersectionPoint`? I kind of like how it reads, I think it makes it easier to guess the result type from the function name, and for two overlapping line segments it seems a little less weird to say "the `intersectionPoint` is `Nothing`" instead of "the `intersection` is `Nothing`".

I think this would probably mean renaming some of the other (future) homogeneous intersection functions to `Plane3d.intersectionAxis`, `Triangle3d.intersectionLineSegment` etc., and keep the generic `intersection` name for cases where the result is some sort of custom union type. It is more verbose, though, and it's certainly simpler to just use `intersection` as the name for all intersection functions.

mpizenberg:
I think you're right. It's been a long road from the initial draft and at this point (no pun intended) the name `intersectionPoint` makes more sense to me than just `intersection`. We loose indeed the fact that all homogeneous functions have the same name, but each gain readability.

ianmackenzie:
Great, merged! Thanks again for all your work on this, both the implementation work and (every bit as valuable) all the productive discussion. Feel free to open a new pull request if you think of any more improvements/fixes =)

Issue #1, opened 2016-10-24
Add Interval type and module


ianmackenzie:
Will be used for return value of functions like `LineSegment3d.distanceAlong`.

Type should be `type Interval = Interval { minValue : Float, maxValue : Float }`.

Functions should include:
- [x] `singleton : Float -> Interval`
- [x] `minValue : Interval -> Float`
- [x] `maxValue : Interval -> Float`
- [x] `width : Interval -> Float`
- [x] `midpoint : Interval -> Float`
- [x] `contains : Float -> Interval -> Bool`
- [x] `intersects : Interval -> Interval -> Bool`
- [x] `isContainedIn : Interval -> Interval -> Bool`
- [x] `hull : Interval -> Interval -> Interval`
- [x] `hullOf : List Interval -> Maybe Interval`
- [x] `intersection : Interval -> Interval -> Maybe Interval`
- [ ] `sin : Interval -> Interval`, `cos : Interval -> Interval` (useful for implementing `Arc2d.boundingBox`)

ianmackenzie:
Largely complete via 2bcdd2974e99e15eea5255bfd86f6abde23dfc2d and 13aacafd1d7d0c57a3eb787252be5c10fde85a30; `sin`/`cos` moved to new issue #26.

PR #21, opened 2017-09-19
Arc Length & Length Parameterization 

folkertdev:
This PR aims to implement length parameterization 

For curves without a direct solution for the length, we add 

```elm
approximateLength : Float -> a -> Float
arcLengthParameterization : Float -> a -> (Float -> Maybe (Point2d))
```
where the first argument is the maximum error percentage. 

## method 

The curve is approximated based on a config: 

```elm
type alias LengthConfig a =
    { split : Float -> a -> ( a, a )
    , percentageError : Float
    , startAndEndpoint : a -> ( Point2d, Point2d )
    }
```
using the config, the curve is recursively approximated by straight line segments until the difference in lengths between the current iteration and the previous iteration is below percentageError. 

The above gives a polyline. Length can be calculated on the polyline, and the polyline can be used for the length parameterization

To make the lenght parameterization efficient, we build a tree where the leaves are line segments, and the nodes contain the length of their left child and the sum of the lengths of the children. The tree auto-balances, code based on [elm-tree by Ilias](http://package.elm-lang.org/packages/zwilias/elm-tree/1.0.0/Tree-AVL)

I've verified that when applying the first two arguments, the tree is constructed once. So in the below snippet, the tree is only constructed once and reused for both invocations of `f`. 
```elm
arcLengthParameterization : Float -> a -> (Float -> Maybe (Point2d))

f = arcLengthParameterization 0.001 mySpline

f 0
f 1 
```

## open issues 

* This test should pass, but does not at the moment because of float errors. 

```elm
           Test.fuzz Fuzz.quadraticSpline2d "at s = (length spline) the arcLengthParameterization gives the end point" <|
               \spline ->
                   let
                       error =
                           1.0e-6
                   in
                       QuadraticSpline2d.approximateLength error spline
                           |> QuadraticSpline2d.arcLengthParameterization error spline
                           |> Expect.equal (Just <| QuadraticSpline2d.endPoint spline)
```

* Clear documentation on what the error means - and verifying that it does what we say it does. Also, we say percentage error, but don't we actually use the "fractional" value? when we want the error to be 1%, is the argument 1 or 0.01?


mpizenberg:
Hi, just a remark about the error in test. I've not read the code, but from your explaination, the failure seems normal. It is not because 2 iterations have a difference of (let's say) less than 0.1 that the actual correct value is at a difference less than 0.1. This depends on the convergence of the sequence.
Let me know if this is not clear of if I misinterpret something.

folkertdev:
my current problem is that the root node "lies" about its children, saying that they are longer than they actually are. Some Debug.log statements give me 

```
arc length: 5.196588393529261

-- searching the tree ( s, lengthAtSplit, lengthAtEnd)
looking at: (5.196588393529261,2.0872275335282864,5.196588393529261)
looking at: (3.1093608600009746,1.022623275942056,3.109360860000974)

3.1093608600009746
3.109360860000974
```
the arc length is the root's lengthAtEnd  because length and arcLengthParameterization use the same tree right now. Next, root node is saying "ow yes I contain this value", but then the right child actually doesn't.

ianmackenzie:
Quick update: I've started looking at this in the [pr/21 branch](/opensolid/geometry/tree/pr/21) - I've updated the code to use the newly-renamed functions like `fromCoordinates` and `fromVertices` and started to look at the logic.

My current thinking is to construct the tree directly as part of the recursive curve subdivision instead of accumulating a list of segments and then building a tree from them - I think it should then be possible for the tree to store absolute arc length values (instead of the right-hand tree starting from zero), by building up the left tree first and then using its final arc length value as an initial arc length value when constructing the right-hand tree. This should then make it possible to look up what leaf to use for arc length parameterization simply by comparing values, with no subtraction as is currently done in the final case of `evaluateTreeAt` - I think there might be some mismatch between roundoff done during addition when finding the total arc length and roundoff done during subtraction when evaluating at a given arc length.

Another thing I'd like to modify is the termination criterion for the approximation - it seems to me the current logic of looking at relative error between successive refinements can fall down in pathological cases. For example, consider an S-shaped cubic spline where evaluating at t=0.5 gives the midpoint between the start and end points - if the first approximation was a single segment (start to end) and the second approximation was two segments (start to mid, mid to end) then the difference in arc length between the two would be zero and the refinement process would terminate. I'd like to modify the logic to have the `config` record have a field

```elm
arcLengthBounds : a -> { lowerBound : Float, upperBound : Float }
```

where for a spline the lower bound is the length from the start point to the end point and the upper bound is the length of control polygon (sum of lengths between successive control points). Then you can *guarantee* that the arc length error is bounded properly (within floating-point roundoff, anyways).

I'll continue to work on this over the next few days - hopefully have some code to discuss soon!

ianmackenzie:
Side note: I thought the GitHub default was for a PR to create a special branch that I could push to directly, but that doesn't seem to be the case here - @folkertdev, is there a way for you to edit the PR so that this is the case? Then I can merge the `pr/21` branch into the pull request branch and all new commits should show up here in the PR itself...that's what @mpizenberg  and I did in #17.

ianmackenzie:
A couple more points for discussion:

### Interpolation along curve

Currently evaluation works by searching the tree for an appropriate approximation line segment and interpolating along it. It occurs to me that it would be better to store pairs of start/end parameter values at each tree leaf instead of a line segment, then interpolate between those two parameter values and evaluate the original spline at the interpolated parameter value. This would mean that all evaluated 
points would actually be on the original spline instead of on its polyline approximation.

### Tangent direction evaluation

I think in many cases it would be useful to also be able to evaluate the tangent direction (as well as the position) at a given arc length. That implies that `arcLengthParameterization` should perhaps return a record with type

```elm
{ pointAt : Float -> Maybe Point2d
, tangentDirection : Float -> Maybe Direction2d
, evaluate : Float -> Maybe ( Point2d, Direction2d )
, arcLength : Float
}
```

(This would also mean you could avoid building the tree twice, by just calling `arcLengthParameterization` and using the `arcLength` field instead of calling `approximateLength` separately.)

ianmackenzie:
Side note for anyone following along: I was eventually able to commit to the `folkertdev:master` branch, although I needed to run

```
git config lfs.https://github.com/folkertdev/geometry.git/info/lfs.locksverify false
```

in my local checkout of the branch (as discussed [here](https://github.com/git-lfs/git-lfs/issues/2190)) in order to be able to push.

folkertdev:
* build up tree as you go: 
sounds good. we'll have to make sure the rebalancing of the tree does not introduce errors. 

* lowerbound/upperbound: 
I had this in an earlier version, and it makes a lot of sense in the example you described. 

* storing the parameter at the leaf: 
very interesting idea. I think we'd also still need to store the length/distance information in the leaf as a source of truth when rebalancing. 

* tangent direction: 
I played around with the proposed returned record a bit and I like it. 
Would it be useful to give the parameter value for some length, possibly instead of the tangent? 


Based on the above, I think this is the tree type that follows: 

```elm
type SegmentTree 
    = Node { lengthAtStart : Float, lengthAtEnd : Float } Int SegmentTree SegmentTree 
    | Leaf { lenghtAtStart : Float, paramAtStart : Float, lengthAtEnd : Float, paramAtEnd : Float } 
```

ianmackenzie:
> - build up tree as you go:
      sounds good. we'll have to make sure the rebalancing of the tree does not introduce errors.

I'm actually wondering if we can get away without rebalancing at all - I'm struggling to think of a spline where you wouldn't get a reasonably-balanced tree by just recursively bisecting the spline and building a subtree from each half. It wouldn't be perfectly balanced, but I can't think of any truly pathological cases (anything equivalent to inserting sorted values into a binary tree, a very reasonable input leading to a totally unbalanced tree).

> - tangent direction:
      I played around with the proposed returned record a bit and I like it.
      Would it be useful to give the parameter value for some length, possibly instead of the tangent?

Interesting - so something like a `parameterValue`, `toParameterValue` or `parameterValueAt` function field? That would certainly allow the most flexibility.

folkertdev:
Thinking one step further ahead: What if I have a list of bezier segments and want 10 evenly spaced points along the curve that they form? (this is what I ultimately want to use the length parameterization for in `one-true-path`)

I think that would be easiest and most efficient if the `SegmentTree`s of the different parts could be merged into one, so `SegmentTree` would have to be exposed too. Then we could have 

```elm
type alias ArcLengthParameterization =
    { arcLength : Float
    , pointAt : Float -> Maybe Point2d
    , tangentAt : Float -> Maybe Direction2d
    , tree : SegmentTree
    }

construct : Config a -> a -> ArcLengthParameterization
union : ArcLengthParameterization -> ArcLengthParameterization -> ArcLengthParameterization
```

The parameter doesn't work in this case because we can't know to which segment the parameter belongs.

ianmackenzie:
Hmm, interesting - although I'm a bit hesitant to expose something low-level like `SegmentTree`. I think in that case it might make more sense to have an opaque `ArcLengthParameterization` type, and have functions to construct and act on values of that type:

```elm
arcLengthParameterization : Float -> QuadraticSpline2d -> ArcLengthParameterization
arcLength : ArcLengthParameterization -> Float
pointAtLength : Float -> ArcLengthParameterization -> Maybe Point2d
tangentDirectionAtLength : Float -> ArcLengthParameterization -> Maybe Direction2d
evaluateAtLength : Float -> ArcLengthParameterization -> Maybe ( Point2d, Direction2d )
concat : List ArcLengthParameterization -> ArcLengthParameterization
```

(I like `concat` because the behavior is pretty similar to `List.concat`/`String.concat` in that it concatenates things together in order, as opposed to something like `Set.union`.) An alternative to `concat` would be to just support constructing an arc length parameterization from a list of splines:

```elm
piecewiseArcLengthParameterization : Float -> List QuadraticSpline2d -> ArcLengthParameterization
```

I like this since I think in the future there will be functions like `CubicSpline2d.catmullRom` that return a list of splines, which you could then pipe directly into `piecewiseArcLengthParameterization`. If we think this is the common case we could even make it the default:

```elm
arcLengthParameterization : Float -> List QuadraticSpline2d -> ArcLengthParameterization
```

folkertdev:
that the second part won't quite work for `one-true-path`, because there the segments won't all be of the same type. I could have 

```
mySegments = [ LineSegment p1 p2, LineSegment p2 p3, CubicSegment p3 c1 c2 p4 ] 
```
and I'd want (say) evenly spaced points over the whole curve. 

I like the idea of an opaque ArcLengthParameterization. The API you give is nice, but I think I'd make the `ArcLengthParameterization` the first argument:

```
List.map (pointAtLength myParameterization) [ 1,2,3 ] 
-- vs 
List.map (flip pointAtLength myParameterization) [ 1,2,3 ] 
```
I think it's more likely to check many lengths on one parameterization than one length on many parameterizations.

ianmackenzie:
Agreed on the argument order.

> the second part won't quite work for one-true-path, because there the segments won't all be of the same type

How would a separate `union`/`concat` work in that case? It seems then `ArcLengthParameterization2d` would need to be a top-level type and there would need to be `arcLengthParameterization` functions for `LineSegment2d` and `Arc2d` as well as for the spline types.

I should mention that I'm currently working on a polymorphic `Curve2d` type in [`opensolid/parametric`](https://github.com/opensolid/parametric) - it might actually make more sense to have all these functions in the `Curve2d` module instead of `QuadraticSpline2d` etc., since then you would be able to create and merge parameterizations for different curve types. But I'd guess I'm at least a month or two away from an initial release of that package.

Might it make sense to implement a non-mergeable monomorphic `QuadraticSpline2d.ArcLengthParameterization` type (and same for the other spline types) for now, then perhaps in the future add a generic `Curve2d.ArcLengthParameterization` type that *can* be merged? In the meantime the merged arc length parameterization for `one-true-path` should be possible with a simple linear search through a `one-true-path`-specific list of things like

```elm
type ArcLengthParameterization
    = QuadraticSplineParameterization OpenSolid.QuadraticSpline2d.ArcLengthParameterization
    | CubicSplineParameterization OpenSolid.CubicSpline2d.ArcLengthParameterization
    | ArcParameterization Arc2d Float
    | LineSegmentParameterization LineSegment2d Float
``` 

where the extra `Float` parameters just cache the length of the arc or line segment. That should be fast unless there are a *lot* of path segments...

ianmackenzie:
I've just committed an initial working version - see what you think. I'm not totally sure about the API yet but I think it's getting there.

I tried out a different approach for tolerance where the given value is treated as an absolute tolerance on the total arc length, and then a simple (stupid?) approach is used of allocating half of that tolerance to each side when bisecting. We could certainly play with this to try out different approaches (I was pondering having the top-level given tolerance be an absolute value, but then immediately convert it into a relative value to use internally by dividing it by the lower bound on the overall arc length, i.e. the distance between the curve endpoints).

folkertdev:
That's great!

I think the approach you mention for `one-true-path` will work fine for now. 

Having an opaque ArcLengthParamterization per module makes sense given that there will one day be a generic way. 

For testing the error, it would be nice to have some non-trivial bezier curves of which we know the exact length. Only then can we actually evaluate how the error behaves and what the tradeoffs are. Can't really find any examples of bezier curves with their exact lengths though.

mpizenberg:
Let's consider a quadratic bezier whose control points are A, B, C.
Let f be the curve function.
For t in [0,1], f(t) = (1-t)^2 A + 2t(1-t) B + t^2 C
The length of the curve is L = integral_A^C( ds )
where ds = sqrt( (dfx/dt)^2 + (dfy/dt)^2 )
dfx/dt = -2(1-t) Ax + 2(1-2t) Bx + 2t Cx
by developing,
dfx/dt = alpha_x + beta_x t
where
alpha_x = 2(B-A)x
beta_x = 2(A-2B+C)x

(dfx/dt)^2 = alpha_x^2 + 2 alpha_x beta_x t + beta_x^2 t^2

same with y, so
(dfx/dt)^2 + (dfy/dt)^2 = |alpha|^2 + 2 alpha*beta t + |beta|^2 t^2

I'm a bit tired to continue for now but seems that now we have an analytic formula
with the help of this answer: https://stackoverflow.com/questions/11854907/calculate-the-length-of-a-segment-of-a-quadratic-bezier
where their A, B, and C are: (sorry for calling the initial points A, B and C, I hope that will not introduce confusion thanks to intermediary variables alpha and beta ...)
A := |beta|^2
B := 2 alpha*beta
C := |alpha|^2

I hope I didn't make a mistake. The analytic formula could enable the verification of any test example :)

mpizenberg:
oops, I read "bezier" and "quadratic" and I interpolated (no pun intended). But I don't see quadratic bezier curve in the types so I guess you meant quadratic spline @folkertdev ?

folkertdev:
yes, though I guess we could just take a circle segment as the test case, where it it easy to calculate the exact length.

mpizenberg:
If I remember, it's quite dificult to represent arcs of circle with splines, and that would be the reason why there are also non-uniform rational b-spline (nurbs), a bit like how projections are done using an additional coordinate in projective coordinates.

folkertdev:
The arcLengthParameterization is now implemented for quadratic splines, but we can easily also write implementations for other sorts of geometry. Then we can test whether the approximations make sense and experiment with error/tolerance propagation strategies on a piece of geometry where it is easy to verify correctness - for instance circle segments.

mpizenberg:
Hi, so I've been re-reading a bit the QuadraticSpline2d module. Turns out it is (as said in the doc ^^) a bezier curve, and not a natural uniform spline like my mind thought at second thought. So the analytic formula was the right thing to compute :).

I've redone it more properly (cf [joint image][curve_length]) et written an [ellie example][ellie] that computes what I've manually obtained. However it appears that I still have an error somewhere since the length computed is incorrect (the ellie code use the example of a line).

Let me know if you think that's interresting and you can spot the mistake.

[ellie]: https://ellie-app.com/rGrzG6Q3Ka1/0
[curve_length]: https://user-images.githubusercontent.com/2905865/30816091-8a6f72ac-a247-11e7-9e38-f6fd3930dc6d.jpg

ianmackenzie:
Looks like the arc length computation is working correctly! (See test added in the last commit.)

@mpizenberg, when you perform your reparameterization, shouldn't the ||A||^4 term actually be ||A||^2||B||^2 so that ||A||^2||B||^2 / ||B||^4 cancels out and recovers the original ||A||^2 / ||B||^2 term?

I think this is getting close to ready to merge, although I'd like to mull over the API a bit more first.

mpizenberg:
@ianmackenzie You are totally right, I squared instead of multiplying by ||B||^2 !
Thank you! There is the updated (working woohoo) [ellie example][ellie]

I think the advantage of this formulation over the one you used in the test is that it works every time (event for degenerate - line - splines) EXCEPT when the second control point is the mid point of the first and last control point.

[ellie]: https://ellie-app.com/rGrzG6Q3Ka1/1

ianmackenzie:
@mpizenberg great! I may switch over to your formulation at some point - it would be great to just work on any spline, perhaps with a special case for the "pure" line case where the second control point is the midpoint.

mpizenberg:
Hi, just a small remark about the last analytic form I sent. I fear it might be unstable (how much is to be evaluated) when the mid control point converges to mid of the segment since |B| -> 0 and:

`gamma` is in |B|^(-2)
`beta` in in |B|^(-2)
`k` is thus in |B|^(-4)

by developing `k` as `gamma - beta^2`, we can get rid of the |B|^(-4) terms but I can't figure out a way to get rid of the |B|^(-2) terms.

I wonder if there is a way to rewrite this expression in such a way that is stable when p1 converges to (p0+p2)/2. I couldn't figure it out. Anyway, I just thought I should drop the thought here.

ianmackenzie:
@mpizenberg I'm not too concerned - if this was exposed as `QuadraticSpline2d.arcLength` that would be one thing, but as just a check to make sure the arc length parameterization is working properly it's not a big deal. If necessary we could add a separate test to check linear splines...

ianmackenzie:
I was playing around with some ideas for a tweaked API that would be a bit more consistent with the rest of `opensolid/geometry` (and a bit more flexible, too). First, add an exposed `ArcLength` module with an opaque `Parameterization` type (actually an alias to the existing non-exposed `ArcLengthParameterization`):

```elm
module OpenSolid.ArcLength
    exposing
        ( Parameterization
        , toParameterValue
        , fromParameterValue
        )

type alias Parameterization =
    ArcLengthParameterization.ArcLengthParameterization

toParameterValue : Parameterization -> Float -> Maybe Float

fromParameterValue : Parameterization -> Float -> Maybe Float
```

(It seems pretty simple to support `fromParameterValue` with a slight modification to the existing segment tree structure, it's nicely symmetric with `toParameterValue`, and I can see it potentially being useful.) The relevant bits of the `QuadraticSpline2d` and similar modules would then look something like

```elm
arcLengthParameterization : Float -> QuadraticSpline2d -> ( Float, ArcLength.Parameterization )

pointAlong : QuadraticSpline2d -> ArcLength.Parameterization -> Float -> Maybe Point2d

tangentAlong : QuadraticSpline2d -> ArcLength.Parameterization -> Float -> Maybe Direction2d
```

I think I like returning arc length directly from `arcLengthParameterization`, since I can't think of any cases where you could use the returned parameterization *without* getting the arc length first, and I like how it reads:

```elm
( arcLength, parameterization ) =
    QuadraticSpline2d.arcLengthParameterization 0.001 spline
```

The evaluation functions would then just be convenient shorthand for combining parameterization and evaluation; for example,

```
QuadraticSpline2d.pointAlong spline parameterization length
```

would be equivalent to

```
ArcLength.toParameterValue parameterization length
    |> Maybe.map (QuadraticSpline2d.pointOn spline)
```

The `*Along` function names are perhaps a bit more obscure than the `*AtLength` versions, but I like how they follow the common pattern of "function name refers to the first argument" (for example `pointAlong` refers to the spline the point is along as its first argument, as opposed to the existing `pointAtLength` where the length is the second argument). And they follow the naming pattern of other functions like `Point2d.along`, `Point3d.distanceAlong`, `Frame3d.translateAlongOwn` which also deal with distances along something.

Thoughts? One risk is that you could call `pointAlong` with a spline and the parameterization for a totally different spline, but in return you get a nice conceptual decoupling between the spline itself and its arc length parameterization.

folkertdev:
you can regain type safety by using phantom types. leads to pretty long type signatures but it's safe.

I suspect that when using many types and their parameterizations (as I will with my segments) it could happen that I call a function with the wrong type argument. 

```elm
type alias Parameterization a =
    ArcLengthParameterization.ArcLengthParameterization

toParameterValue : Parameterization a -> Float -> Maybe Float

fromParameterValue : Parameterization a -> Float -> Maybe Float

--

arcLengthParameterization : Float -> QuadraticSpline2d -> ( Float, ArcLength.Parameterization QuadraticSpline2d )

-- etc

```

on *Along vs. *AtLength: I like *AtLength better, but along works as well, your consistency argument makes sense.

ianmackenzie:
Phantom types would achieve type safety but not 'value safety'...you could still use a `Parameterization QuadraticSpline2d` with a different `QuadraticSpline2d` than was used to create the parameterization initially.

Here's a type safe variation that is even closer to what we have now:

```elm
module OpenSolid.ArcLength
    exposing
        ( Parameterized
        , toParameterValue
        , fromParameterValue
        )

import ArcLengthParameterization

type alias Parameterized curve =
    ArcLengthParameterization.Parameterized curve

toParameterValue : Parameterized curve -> Float -> Maybe Float

fromParameterValue : Parameterized curve -> Float -> Maybe Float
```

(This would require a slight modification to the `ArcLengthParameterization` module to additionally store the underlying curve.) `QuadraticSpline2d` would then look like

```elm
module OpenSolid.QuadraticSpline2d

arcLengthParameterization : Float -> QuadraticSpline2d -> ( Float, ArcLength.Parameterized QuadraticSpline2d )

pointAlong : ArcLength.Parameterized QuadraticSpline2d -> Float -> Maybe Point2d

tangentAlong : ArcLength.Parameterized QuadraticSpline2d -> Float -> Maybe Direction2d
```

I kind of like how it reads: "a point along an arc length parameterized quadratic spline". In some ways this is a bit more complex than the `ArcLength.Parameterization` version due to the generic `Parameterized` type, but I think "a curve parameterized by arc length" is also a slightly more concrete and understandable concept than "an arc length parameterization".

mpizenberg:
I also like the reading, but if curve can be any type, how to implement
*ParameterValue ?

mpizenberg:
Forget what I just said, hard to think at wake up ^^

ianmackenzie:
The `ArcLength` module itself wouldn't know anything about different curve types, but once you've built up the segment tree (in exactly the same way we do now, using a config object), it contains enough information to go back and forth between length and parameter values without knowing anything about the original curve. The actual curve would only be stored in the `Parameterized` value so that the various different curve modules could extract it later to evaluate points, tangents etc.

ianmackenzie:
So a few more thoughts (sorry for the delay in getting back to this):

## Correctness

I think `lengthBounds` in the config object will have to change to `speedBounds` or similar - bounds on the magnitude of the first derivative, not bounds on the overall length. Without this, `toParameterValue` and `fromParameterValue` could give incorrect results. For example, if we used the analytic formula for quadratic spline length, then `lengthBounds` could just return the exact length value for the upper and lower bound and tree construction would terminate immediately with a single leaf. Then basically arc length parameterization would end up being just the same as normal parameterization! By asking for bounds on magnitude of the first derivative I think we can make this correct in all cases.

## Generalization

I'm trying to figure out how this will expand to the polymorphic `Curve2d` and `Curve3d` types, which are (at least currently) in a separate package - if there's a non-exposed module in `opensolid/geometry` that implements actual construction of an `ArcLengthParameterized c` then that wouldn't be available to code in another package. Even if there was an exposed function

```elm
module OpenSolid.ArcLength exposing (..)

parameterizedCurve : Config c -> c -> ArcLengthParameterizedCurve c
```

(which would mean exposing the `Config` type publicly) then to build an arc-length parameterized `Curve2d` you'd have to figure out how to construct a `Config Curve2d` - this could delegate to a `Config QuadraticSpline2d` when the curve is a quadratic spline (and use a trivial implementation for arcs and line segments), but that would mean doing something like making the config object public in `QuadraticSpline2d` so that `Curve2d` could use it, which seems messy.

In a lot of ways this goes away if there is a separate `ArcLengthParameterization` type which is independent of any particular curve type, since then `Curve2d.arcLengthParameterization` *can* simply delegate to `QuadraticSpline2d.arcLengthParameterization` etc. But the type safety of embedding the curve type in the parameterization is nice...thoughts? Suggestions?

ianmackenzie:
One possibility - use `ArcLengthParameterized` as a type alias:

```elm
module OpenSolid.ArcLength
    exposing
        ( ArcLengthParameterized
        , ArcLengthParameterization
        , Config
        , parameterized
        , toParameterValue
        , fromParameterValue
        )

-- An arc length parameterized curve is literally
-- just a curve plus its arc length parameterization
type alias ArcLengthParameterized c =
    ( c, ArcLengthParameterization )

type ArcLengthParameterization =
    -- segment tree

type alias Config c =
    { bisect : c -> ( c, c )
    , speed : c -> Interval
    }

parameterization : Config c -> Float -> c -> ArcLengthParameterization

toParameterValue : ArcLengthParameterization -> Float -> Float

fromParameterValue : ArcLengthParameterization -> Float -> Float
```

(If you're wondering about the `Interval` type, it's something I've been meaning to add for a while - see 
#1.) Almost everything outside of `ArcLength` would just accept and return `ArcLengthParameterized` values:

```elm
module OpenSolid.QuadraticSpline2d

import ArcLength exposing (ArcLengthParameterized)

arcLengthParameterized : Float -> QuadraticSpline2d -> ArcLengthParameterized QuadraticSpline2d

arcLength : ArcLengthParameterized QuadraticSpline2d -> Float

pointAlong : ArcLengthParameterized QuadraticSpline2d -> Float -> Maybe Point2d

tangentAlong : ArcLengthParameterized QuadraticSpline2d -> Float -> Maybe Direction2d
```

However, code that wanted to use `toParameterValue` and `fromParameterValue` directly could destructure the `ArcLengthParameterized` values to get the underlying parameterization, and eventually `Curve2d` and `Curve3d` would be able to arc length parameterize the underlying curve type (quadratic spline etc.) but then just construct a new tuple with themselves as the first argument to end up with a proper `ArcLengthParameterized Curve2d` etc.

(I switched from `ArcLength.Parameterized` to `ArcLengthParameterized` because I remembered that the Elm documentation generator strips off all module prefixes from type names, so `ArcLength.Parameterized` would just show up in the documentation type signatures as `Parameterized`, which is a bit vague.)

src/OpenSolid/QuadraticSpline2d.elm
```diff
@@ -460,59 +461,78 @@ splitAt t spline =
...
 
-        lowerBound =
-            Point2d.distanceFrom p1 p3
+        b =
+            sqrt bSquared
 
-        upperBound =
-            Point2d.distanceFrom p1 p2 + Point2d.distanceFrom p2 p3
+        minValue =
+            if cSquared == 0 then
```
mpizenberg: Is this `speed` the norm of the derivative ? if so, I think in case `vC == 0` it means it is a round trip with same starting and ending points? And in that case the derivative nullifies at half course right? Should it be 0 then?

mpizenberg: Oops, ok `vC = vA - vB` not `+` so it is not the round trip but the degenerate segment. Got it!

folkertdev:
Really liking these most recent changes. It's great that we can now define custom 
parameterizations. 

I'm probably missing some information here, but what is the intuition behind the name "speed"?

ianmackenzie:
> what is the intuition behind the name "speed"?

It's the speed at which you would be travelling along the curve, if the rate of change of the parameter value was 1 unit per second - the rate of change of arc length with respect to parameter value. I fully agree that it is not a perfect name, but I was trying to think of something a bit less wordy than `firstDerivativeMagnitude` - certainly open to suggestions, and maybe `firstDerivativeMagnitude` (or the even more explicit `firstDerivativeMagnitudeBounds`) is actually the right way to go.

folkertdev:
What's the status here? Is the API satisfactory now and can we proceed to implement this for other types?

ianmackenzie:
Just implemented and tested arc length parameterization for `QuadraticSpline3d` and `CubicSpline2d`. I also tried out `ArcLengthParameterized` being a proper union type, which I think works pretty well, and renamed `speed` to `derivativeMagnitudeBounds` to be extra explicit.

Accurately determining min/max bounds on cubic spline derivatives is trickier than I thought - it seems to be working for `CubicSpline2d` (I added a dedicated test just to check that derivative magnitude bounds are being calculated correctly), but will require a bit more work for `CubicSpline3d`.

mpizenberg:
So I was curious about the subdivision implementation since you have trouble for the 3D case so I just peaked at it this morning. Please correct me if I'm saying something wrong.

* It is a depth first implementation, meaning we bisect over parameter and compute left branch until it converges to a leaf. Meaning we now have a "correct" left length to initialize the length at start of right branch
* We half the tolerance for each subdivision so that in total, the sum of leafs tolerances equals the initial tolerance setting.


Regarding the difficulties with derivative magnitudes, I wonder it if is worth the trouble. It is only used to estimate the errors on the lenght right? If so one method far less computing intensive (but less precise) could be to estimate it as `0.5 * (|p3-p2| + |p2-p1| + |p1-p0| - |p0-p3|)` if you estimate the length at mean of straight line and contour.

ianmackenzie:
@mpizenberg you're exactly correct about the tree construction, although I'd like to take a look at making it a bit more clever (perhaps using estimated lengths of each bisected segment to determine how much tolerance to allocate to each segment instead of just allocating half the tolerance to each). 

However, I don't think simply placing bounds on the overall length of each segment is sufficient - as far as I can tell you really have to place bounds on the linearity of arc length with respect to parameter value, which is what you get from bounds on derivative magnitude.

For example, consider a quadratic spline with control points (0, 0), (1, 0) and (10, 0). The straight line and contour lengths are both exactly 10, but if you linearly interpolate parameter values within that spline you will not get anything close to a proper arc length parameterization!

One solution is to interpolate linearly between start and end points once the overall length bounds converge to a certain tolerance, but then calling `pointAlong` will not actually give you points *on* the spline. For example, consider a quadratic spline with control points (0, 0), (5, 1) and (20, 0); the lower bound on the length is 20 and the upper bound is 20.13, but if you interpolate linearly between the start and end points you'll get (5, 0) instead of (5, 1).

mpizenberg:
> making it a bit more clever (perhaps using estimated lengths of each bisected segment ...

I was thinking of something similar but then I checked what happens for example when 2 of the 3 control points are the same. In that case, which is probably not far from worst case (I don't know actually), the lengths at parameter 0.5 are split in 1/4 and 3/4 for each side. One side is 3 times longer, but that's not a huge deal. It may probably "only" cost one additional tree depth level for the longer side, so it is mainly a matter of performances rather than precision.

> For example, consider a quadratic spline with control points (0, 0), (1, 0) and (10, 0). The straight line and contour lengths are both exactly 10, but if you linearly interpolate parameter values within that spline you will not get anything close to a proper arc length parameterization!

Ahah yes! I forgot that we were trying to build a parameterization :) not just compute length ...

ianmackenzie:
Just finished an initial working implementation of `CubicSpline3d` arc length parameterization, so I think this is pretty much ready to merge. It's not yet as fast as I would like, but I think everything works. Future pull requests to improve the efficiency of `Internal.CubicSpline3d.derivativeMagnitudeBounds` would definitely be welcome!

ianmackenzie:
I think the performance improvements are now done, which have made things much faster (1 to 2 orders or magnitude). I've added an actual benchmark program in `benchmarks/` to help with future performance improvement work.

I confess I'm having some second thoughts about `ArcLengthParameterized` - every API I can think of also requires `ArcLengthParameterization` to also be exposed, which leads to more concepts to keep track of and some ambiguity about what kind of arguments different functions take (e.g. should `ArcLength.toParameterValue` take an `ArcLengthParameterized c` or an `ArcLengthParameterization`?). @folkertdev how strongly do you feel about `ArcLengthParameterized`? I'm tempted to go back to something simpler, similar to what I proposed earlier:

```elm
module OpenSolid.ArcLength

type ArcLengthParameterization

type alias Config =
    { tolerance : Float
    , derivativeMagnitude : Float -> Float
    , maxSecondDerivativeMagnitude : Float
    }

parameterization : Config -> ( Float, ArcLengthParameterization )
toParameterValue : ArcLengthParameterization -> Float -> Maybe Float
fromParameterValue : ArcLengthParameterization -> Float -> Maybe Float
```

```elm
module OpenSolid.QuadraticSpline2d

arcLengthParameterization : Float -> QuadraticSpline2d -> ( Float, ArcLengthParameterization )
pointAlong : QuadraticSpline2d -> ArcLengthParameterization -> Float -> Maybe Point2d
tangentAlong : QuadraticSpline2d -> ArcLengthParameterization -> Float -> Maybe Direction2d
```

folkertdev:
The problem that brings back is that for instance the `QuadraticSpline2d` pointAlong could be called 
with a `QuadraticSpline3d` curve/parameterization, giving nonsensical results. Actually, 
the types you propose above also allow to use multiple `QuadraticSpline2d` on just one parameterization, giving nonsensical results. 

I think the better route is to (only) keep `ArcLengthParameterized a`, ensuring that only sensible functions can be applied and preventing accidental re-use of a parameterization on a different curve. 

```elm
module OpenSolid.ArcLength

type ArcLengthParameterized a = ... 

type Config curve =
  Config 
        { tolerance : Float
        , derivativeMagnitude : Float -> Float
        , maxSecondDerivativeMagnitude : Float
        , curve :: curve
        }

parameterization : Config curve -> ( Float, ArcLengthParameterized curve )
toParameterValue : ArcLengthParameterized curve -> Float -> Maybe Float
fromParameterValue : ArcLengthParameterized curve -> Float -> Maybe Float
```

```elm
module OpenSolid.QuadraticSpline2d

arcLengthParameterization : Float -> QuadraticSpline2d -> ( Float, ArcLengthParameterized QuadraticSpline2d )
pointAlong : ArcLengthParameterized QuadraticSpline2d -> Float -> Maybe Point2d
tangentAlong : ArcLengthParameterized QuadraticSpline2d -> Float -> Maybe Direction2d
```

ianmackenzie:
Just pushed a new version, see what you think! This has an `ArcLength.Parameterization` type as well as `QuadraticSpline2d.ArcLengthParameterized`, `CubicSpline3d.ArcLengthParameterized` etc., but the key new feature is that (e.g.) the `CubicSpline3d` arc length parameterization API is entirely self-contained - it doesn't publicly refer to any of the types from the `ArcLength` module (only uses them internally), so it's possible to understand and use the `CubicSpline3d` module without ever having to look at the `ArcLength` module. I think this decouples the modules nicely and generally makes things a lot easier to understand. (You really only have to be aware of the `ArcLength` module at all if you're actually implementing a new curve type.)

(I used to think this wouldn't work because the eventual `Curve2d` would need to be able to parameterize the underlying curve and then destructure the resulting `ArcLengthParameterized` value to get the actual `ArcLengthParameterization`, but now that it's possible to build an arc length parameterization of any spline 'from the outside' using only publicly exposed functions, I don't think this is an issue any more.)

folkertdev:
Awesome! That looks solid and covers my type safety concerns.

ianmackenzie:
Great! I might mull this over for a day or two to see if anything else occurs to me, but otherwise I think this might actually be ready to merge =)

Actually, now that I think of it, it's possible that `opensolid/geometry` might not need to expose the `ArcLength` module at all - a future `Curve3d.ArcLengthParameterized` type could just be a wrapper around the various concrete `ArcLengthParameterized` types:

```elm
module Curve3d exposing (Curve3d, ArcLengthParameterized, ...)

type ArcLengthParameterized
    = ParameterizedLineSegment Float LineSegment3d -- trivial, just pre-compute the length
    | ParameterizedArc Float Arc3d -- trivial, just pre-compute the length
    | ParameterizedQuadraticSpline QuadraticSpline3d.ArcLengthParameterized
    | ParameterizedCubicSpline CubicSpline3d.ArcLengthParameterized
```

ianmackenzie:
Merged - thanks @folkertdev and @mpizenberg for all your work!

@folkertdev let me know if you're looking to publish a new release of `one-true-path` or anything and are blocked on having a new OpenSolid release with this in. I could publish a point release soon if necessary, but all else equal I'll probably wait a bit and try to merge in the other currently active pull requests and try to add elliptical arcs as well.

folkertdev:
Awesome! 

Yea I want to give the same interface for all curve types, so I'll wait for elliptical arcs. 
If I can be of service anywhere let me know.

ianmackenzie:
Thanks! I may very well want to bounce some API ideas off you once I start in on elliptical arcs - I'll see if I can get started on that this weekend.

PR #23, opened 2017-10-04
Add BoundingBox2d.strictlyOverlaps

tomekwi:
Hello hello 👋

Following up on our quick thought exchange on Slack, I figured I’d propose a new function checking whether or not a bounding box overlaps another strictly – i.e. shares an area with the other, as opposed to just sharing an edge or corner.

This is useful for many collision detection scenarios, where you don’t care whether or not one object is placed next to another, but don’t want to allow one to penetrate the other.

For my use case (which more or less matches the scenario that I described above), calculating this in application code was possible through a combination of `BoundingBox2d.intersection`. But the code would be far from elegant or idiomatic.

### DIFF
--- a/src/OpenSolid/BoundingBox2d.elm
+++ b/src/OpenSolid/BoundingBox2d.elm
@@ -430,14 +430,15 @@ alwaysFalse firstBox secondBox =
 
 
 {-| Check if one box overlaps another by less than, greater than or equal to a
-given amount. You could perform a tolerant collision check (one that only
-returns true only if the boxes overlap by at least some small finite amount,
-and ignores boxes that just barely touch each other) as
+given amount. For example, you could perform a tolerant collision check (one
+that only returns true only if the boxes overlap by at least some small finite
+amount, and ignores boxes that just barely touch each other) as
 
     boxesCollide =
         BoundingBox2d.overlappingBy GT 0.001 box1 box2
 
-(The [`Order`](http://package.elm-lang.org/packages/elm-lang/core/latest/Basics#Order)
+This can be read as "`box1` and `box2` are overlapping by greater than 0.001
+units". (The [`Order`](http://package.elm-lang.org/packages/elm-lang/core/latest/Basics#Order)
 type and its three values `LT`, `GT` and `EQ` are defined in Elm's `Basics`
 module so are available by default in any Elm program.)
 
@@ -530,14 +531,16 @@ overlappingBy order tolerance =
 
 
 {-| Check if one box is separated from another by less than, greater than or
-equal to a given amount. If you were performing clash detection between some
-objects, you could use `separatedBy` on those objects' bounding boxes as a quick
-check to see if the objects had a gap of at least 1 cm between them:
+equal to a given amount. For example, if you were performing clash detection
+between some objects, you could use `separatedBy` on those objects' bounding
+boxes as a quick check to see if the objects had a gap of at least 1 cm between
+them:
 
     safelySeparated =
         BoundingBox2d.separatedBy GT 0.01 box1 box2
 
-(The [`Order`](http://package.elm-lang.org/packages/elm-lang/core/latest/Basics#Order)
+This can be read as "`box1` and `box2` are separated by greater than 0.01
+units". (The [`Order`](http://package.elm-lang.org/packages/elm-lang/core/latest/Basics#Order)
 type and its three values `LT`, `GT` and `EQ` are defined in Elm's `Basics`
 module so are available by default in any Elm program.)
 
--- a/src/OpenSolid/BoundingBox2d.elm
+++ b/src/OpenSolid/BoundingBox2d.elm
@@ -431,7 +431,7 @@ alwaysFalse firstBox secondBox =
 
 {-| Check if one box overlaps another by less than, greater than or equal to a
 given amount. For example, you could perform a tolerant collision check (one
-that only returns true only if the boxes overlap by at least some small finite
+that only returns true if the boxes overlap by at least some small finite
 amount, and ignores boxes that just barely touch each other) as
 
     boxesCollide =
--- a/src/OpenSolid/BoundingBox2d.elm
+++ b/src/OpenSolid/BoundingBox2d.elm
@@ -453,9 +453,9 @@ negative number.
 
 ### Less than
 
-  - `overlappingBy LT 1e-3` will return true if the two boxes overlap by less than
-    0.001 units or if they do not overlap at all (false if they overlap by more
-    than 0.001 units).
+  - `overlappingBy LT 1e-3` will return true if the two boxes overlap by less
+    than 0.001 units or if they do not overlap at all (false if they overlap by
+    more than 0.001 units).
   - `overlappingBy LT 0` will return true only if the two boxes don't touch or
     overlap at all.
   - `overlappingBy LT -1e-3` will always return false! If you care about _how
--- a/src/OpenSolid/BoundingBox2d.elm
+++ b/src/OpenSolid/BoundingBox2d.elm
@@ -447,8 +447,8 @@ boxes.
 
 Boxes that just touch are considered to have an overlap of zero, which is
 distinct from 'no overlap'. Boxes that do not touch or overlap at all are
-considered to have an overlap which is less than zero but not greater or less
-than any negative number.
+considered to have an overlap which is less than zero but not comparable to any
+negative number.
 
 
 ### Less than
@@ -547,7 +547,7 @@ not touch.
 
 Boxes that just touch are considered to have a separation of zero, which is
 distinct from 'no separation'. 'No separation' (overlap) is considered to be
-less than zero but not greater or less than any negative number.
+less than zero but not comparable to any negative number.
 
 
 ### Less than
--- a/src/OpenSolid/BoundingBox2d.elm
+++ b/src/OpenSolid/BoundingBox2d.elm
@@ -500,6 +500,36 @@ given bounding boxes. If the given boxes do not intersect, returns `Nothing`.
     BoundingBox2d.intersection firstBox thirdBox
     --> Nothing
 
+If two boxes just touch along an edge or at a corner, they are still considered
+to have an intersection, even though that intersection will have zero area (at
+least one of its dimensions will be zero):
+
+    firstBox =
+        BoundingBox2d.with
+            { minX = 0
+            , maxX = 1
+            , minY = 0
+            , maxY = 2
+            }
+
+    secondBox =
+        BoundingBox2d.with
+            { minX = 1
+            , maxX = 2
+            , minY = 1
+            , maxY = 3
+            }
+
+    BoundingBox2d.intersection firstBox secondBox
+    --> Just
+    -->     (BoundingBox2d.with
+    -->         { minX = 1
+    -->         , maxX = 1
+    -->         , minY = 1
+    -->         , maxY = 2
+    -->         }
+    -->     )
+
 -}
 intersection : BoundingBox2d -> BoundingBox2d -> Maybe BoundingBox2d
 intersection firstBox secondBox =
--- a/src/OpenSolid/BoundingBox2d.elm
+++ b/src/OpenSolid/BoundingBox2d.elm
@@ -29,6 +29,7 @@ module OpenSolid.BoundingBox2d
         , minY
         , overlaps
         , singleton
+        , strictlyOverlaps
         , with
         )
 
@@ -354,6 +355,14 @@ overlaps other boundingBox =
         && (maxX boundingBox >= minX other)
         && (minY boundingBox <= maxY other)
         && (maxY boundingBox >= minY other)
+
+
+strictlyOverlaps : BoundingBox2d -> BoundingBox2d -> Bool
+strictlyOverlaps other boundingBox =
+    (minX boundingBox < maxX other)
+        && (maxX boundingBox > minX other)
+        && (minY boundingBox < maxY other)
+        && (maxY boundingBox > minY other)
 
 
 {-| Test if the second given bounding box is fully contained within the first
--- a/src/OpenSolid/BoundingBox2d.elm
+++ b/src/OpenSolid/BoundingBox2d.elm
@@ -435,7 +435,7 @@ returns true only if the boxes overlap by at least some small finite amount,
 and ignores boxes that just barely touch each other) as
 
     boxesCollide =
-        BoundingBox2d.overlappingBy GT 0.001 firstBox secondBox
+        BoundingBox2d.overlappingBy GT 0.001 box1 box2
 
 (The [`Order`](http://package.elm-lang.org/packages/elm-lang/core/latest/Basics#Order)
 type and its three values `LT`, `GT` and `EQ` are defined in Elm's `Basics`
@@ -535,7 +535,7 @@ objects, you could use `separatedBy` on those objects' bounding boxes as a quick
 check to see if the objects had a gap of at least 1 cm between them:
 
     safelySeparated =
-        BoundingBox2d.separatedBy GT 1.0e-2 firstBox secondBox
+        BoundingBox2d.separatedBy GT 0.01 box1 box2
 
 (The [`Order`](http://package.elm-lang.org/packages/elm-lang/core/latest/Basics#Order)
 type and its three values `LT`, `GT` and `EQ` are defined in Elm's `Basics`
--- a/src/OpenSolid/BoundingBox2d.elm
+++ b/src/OpenSolid/BoundingBox2d.elm
@@ -20,6 +20,7 @@ module OpenSolid.BoundingBox2d
         , hull
         , hullOf
         , intersection
+        , intersects
         , isContainedIn
         , maxX
         , maxY
@@ -316,7 +317,16 @@ contains point boundingBox =
     (minX <= x && x <= maxX) && (minY <= y && y <= maxY)
 
 
-{-| Test if one bounding box overlaps (touches) another.
+{-| Test if two boxes touch or overlap at all (have any points in common);
+
+    BoundingBox2d.intersects firstBox secondBox
+
+is equivalent to
+
+    BoundingBox2d.intersection firstBox secondBox
+        /= Nothing
+
+but is more efficient.
 
     firstBox =
         BoundingBox2d.with
@@ -342,15 +352,15 @@ contains point boundingBox =
             , maxY = 5
             }
 
-    BoundingBox2d.overlaps firstBox secondBox
+    BoundingBox2d.intersects firstBox secondBox
     --> True
 
-    BoundingBox2d.overlaps firstBox thirdBox
+    BoundingBox2d.intersects firstBox thirdBox
     --> False
 
 -}
-overlaps : BoundingBox2d -> BoundingBox2d -> Bool
-overlaps other boundingBox =
+intersects : BoundingBox2d -> BoundingBox2d -> Bool
+intersects other boundingBox =
     (minX boundingBox <= maxX other)
         && (maxX boundingBox >= minX other)
         && (minY boundingBox <= maxY other)
@@ -363,6 +373,14 @@ strictlyOverlaps other boundingBox =
         && (maxX boundingBox > minX other)
         && (minY boundingBox < maxY other)
         && (maxY boundingBox > minY other)
+{-| DEPRECATED: Alias for `intersects`, kept for compatibility. Use `intersects`
+instead.
+-}
+overlaps : BoundingBox2d -> BoundingBox2d -> Bool
+overlaps =
+    intersects
+
+
 
 
 {-| Test if the second given bounding box is fully contained within the first
@@ -443,7 +461,7 @@ hull firstBox secondBox =
 
 
 {-| Attempt to build a bounding box that contains all points common to both
-given bounding boxes. If the given boxes do not overlap, returns `Nothing`.
+given bounding boxes. If the given boxes do not intersect, returns `Nothing`.
 
     firstBox =
         BoundingBox2d.with
@@ -485,7 +503,7 @@ given bounding boxes. If the given boxes do not overlap, returns `Nothing`.
 -}
 intersection : BoundingBox2d -> BoundingBox2d -> Maybe BoundingBox2d
 intersection firstBox secondBox =
-    if overlaps firstBox secondBox then
+    if intersects firstBox secondBox then
         Just
             (with
                 { minX = max (minX firstBox) (minX secondBox)
--- a/src/OpenSolid/BoundingBox2d.elm
+++ b/src/OpenSolid/BoundingBox2d.elm
@@ -28,8 +28,8 @@ module OpenSolid.BoundingBox2d
         , midY
         , minX
         , minY
+        , overlappingBy
         , overlaps
-        , overlapsBy
         , scaleAbout
         , separatedBy
         , singleton
@@ -71,7 +71,7 @@ box of an object than the object itself, such as:
 
 # Queries
 
-@docs contains, isContainedIn, intersects, overlaps, overlapsBy, separatedBy
+@docs contains, isContainedIn, intersects, overlaps, overlappingBy, separatedBy
 
 
 # Transformations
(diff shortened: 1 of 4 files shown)


tomekwi:
Fun fact:

@ianmackenzie when I described the function to you, you mentioned you’d like to call it `strictlyOverlaps`. That’s exactly the same as I was thinking before I even hit you up!

tests/BoundingBox2d.elm
```diff
@@ -68,6 +69,68 @@ intersectionConsistentWithOverlaps =
...
+                strictlyOverlaps =
+                    BoundingBox2d.strictlyOverlaps first second
+
+                intersection =
+                    BoundingBox2d.intersection first second
+
+                intersectionDimensions =
+                    Maybe.map BoundingBox2d.dimensions intersection
+            in
+                case ( strictlyOverlaps, intersectionDimensions ) of
```
tomekwi: The indentation of the `case` block doesn’t match the rest of the file, apparently because I’m using a different version of elm-format. I’m on `elm-format-0.18 0.6.1-alpha` – should I upgrade/downgrade?

ianmackenzie: Yeah, I'm using 0.7.0-exp. I didn't properly consider the implications on collaboration when I switched to the bleeding-edge version, for which I apologize - in the future I'll be more careful about sticking to the most common/stable version of elm-format. In the meantime, though, if you could use 0.7.0-exp for OpenSolid stuff that would be great.

ianmackenzie:
Hello, and thanks for the PR! =) I agree that this is a use case that should be better supported, although I'd like to think a bit more about the best way to address it - I still feel a bit uneasy about having two functions that are only different in how they treat exact floating-point equality.

One thing I was thinking of (which seems a bit more geometrically meaningful, and more flexible in general) would be have one or two functions that would tell you *how much* two bounding boxes overlapped, or how far apart they were:

```elm
-- How much two boxes overlap: the minimum distance you
-- can move one box so that it no longer overlaps the other
overlap : BoundingBox2d -> BoundingBox2d -> Float

-- How far apart two boxes are: the minimum distance you
-- can move one box to have it touch the other
separation : BoundingBox2d -> BoundingBox2d -> Float
```

Then in your case you'd do something like

```elm
BoundingBox2d.overlap box1 box2 > 0
```

or

```elm
BoundingBox2d.overlap box1 box2 > tolerance
```

For stuff like clash detection, I can see stuff like

```elm
BoundingBox2d.separation box1 box2 > minimumSeparationDistance
```

Thoughts? Not sure at all about naming, I could see the first one being called `penetration` or the second one being called `minimumDistance`...

ianmackenzie:
On a related note, it might also make sense to change the semantics of the existing `overlaps` function to return false for just-touching boxes, and similarly have `intersection` return `Nothing` - I can see that being a more intuitive interpretation, and it's possible that it's more useful in the majority of situations. Would that itself address your use case?

tomekwi:
Hey @ianmackenzie, thanks for sharing your thoughts on this! To address your questions:

> Thoughts? Not sure at all about naming, I could see the first one being called penetration or the second one being called minimumDistance...

TBH I don’t have a strong opinion on how the need should be addressed, but I admit your suggestion seems more sustainable and flexible than what I proposed.

Still thinking about semantics though. Having those two functions return a `Float` rather than something like `Maybe Float` implies that both of them could also return negative values, right? Am I right in thinking that `BoundingBox2d.overlap box1 box2 == -(BoundingBox2d.separation box1 box2)` for any pair of bounding boxes? That might feel like the same kind of redundancy we want to avoid.

Perhaps union types could come to the rescue though? Having a single function like `` that can return things like:

```elm
BoundingBox2d.compare : BoundingBox2d -> BoundingBox2d -> Relationship

type Relationship =
  Overlapping Float
  Touching
  SetApart Float
```


> it might also make sense to change the semantics of the existing overlaps function to return false for just-touching boxes, and similarly have intersection return Nothing - I can see that being a more intuitive interpretation, and it's possible that it's more useful in the majority of situations. Would that itself address your use case?

Yes, that would be awesome! It’s exactly what I’m looking for 😀

Don’t want to be selfish though, so if anyone else (@mpizenberg?) can think or has heard of other use cases which would require `overlaps` to stay eager, I’m happy to discuss and perhaps wait till things like `overlap` and `separation` arrive.

tests/BoundingBox2d.elm
tomekwi: Of course man, my next commit will use the new formatting once we agree on the way forward 👌

mpizenberg:
Hi Tomek, hi Ian,

Let me say first that I'm not sure about anything called `penetration` in OpenSolid ...

[Few months ago][pr17], we've discussed a lot with Ian about segments intersections and came up with the function `intersectionPoint : LineSegment2d -> LineSegment2d -> Maybe Point2d`. The text definition of this function is as follow:

> Attempt to find the unique intersection point of two line segments.
> If there is no such point (the two line segments do not touch,
> or they overlap), returns Nothing.

From the semantic of this documentation, I would conclude that "overlap" has a strict meaning, which is in contradiction with the current behavior of `BBox.overlaps`.

However, the behavior of `intersectionPoint` is consistent with the one of `BBox.intersection` which is a good thing, especially if bounding boxes are to be used generically as a light precondition for checking intersections.

So I would argue that having `BBox.overlaps` and `BBox.intersection` consistent is probably also a good choice with maybe wrong naming since "overlap" in my mind tends to be something strict.

Regarding the rich union type `Relationship`, I'm not sure it fits well with the rest of OpenSolid. This is also my biased opinion since I suggested this also for the [intersection of segments initially][issue4] and we didn't really have a use case where such complexity was useful.
But you seem to have a good use case so I don't know ...

[pr17]: https://github.com/opensolid/geometry/pull/17
[issue4]: https://github.com/opensolid/geometry/issues/4

ianmackenzie:
Thanks both for taking the time to think through this!

@mpizenberg you're exactly right, the current `BoundingBox#d.intersection` function is indeed consistent with `LineSegment2d.intersectionPoint` and the current semantics of `overlaps` are useful for doing things like searching through a quadtree/octree or R-tree (which I definitely intend to implement at some point). I think in general my reasoning was that it was better to be conservative about saying "yes, there's some overlap" or "yes, there's some sort of intersection" since it's easier to filter out 'trivial' intersections from an overly-eager `intersection` etc. than to perform a separate check for them.

> Still thinking about semantics though. Having those two functions return a Float rather than something like Maybe Float implies that both of them could also return negative values, right? Am I right in thinking that BoundingBox2d.overlap box1 box2 == -(BoundingBox2d.separation box1 box2) for any pair of bounding boxes? That might feel like the same kind of redundancy we want to avoid.

@tomekwi I was actually initially thinking that both `overlap` and `separation` would always return nonnegative values - `overlap` would return 'zero overlap' for boxes that were separated by some finite distance, and `separation` would return 'zero separation' for overlapping boxes. Would `overlapWith` and `separationFrom` be any better? The `Maybe Float` versions do seem more clear, but then you can't use them in a simple greater-than or less-than comparison. (If only `Maybe`s were `comparable`!) An alternative might be something like

```elm
overlappingBy : Float -> BoundingBox2d -> BoundingBox2d -> Bool
separatedBy : Float -> BoundingBox2d -> BoundingBox2d -> Bool

-- or if you want to be extra explicit
overlappingByAtLeast : Float -> BoundingBox2d -> BoundingBox2d -> Bool
separatedByAtLeast : Float -> BoundingBox2d -> BoundingBox2d -> Bool
```

so you would write code like

```elm
if BoundingBox2d.overlappingBy overlapTolerance box1 box2 then ....
if BoundingBox2d.separatedBy minimumDistance box1 box2 then ...
```

This does provide a bit more flexibility in implementation (you can do things like compare squared diagonal distances to squared tolerances instead of needing to use a square root to return an actual distance), although then you can't easily check for "any overlap greater than zero" (you'd have to provide a small tolerance - although that may not be a bad idea anyways).

I'm also not sure about a `compare` or `classify` function that would return a union type - my gut feeling is that in most cases you just want to make a binary decision about two bounding boxes as opposed to doing one of several different things depending on the nature of the relationship, so we should try to figure out the important binary decisions people need to make and just support those directly. (Side note: you could argue that a `Relationship` type should also distinguish between cases like 'A and B overlapping', 'A contained in B', 'B contained in A' - I can imagine cases where that would be useful.)

tomekwi:
What you guys are saying makes a lot of sense! I especially like the boolean versions `overlappingBy` `separatedBy`. Seems like the simplest possible way to cover the collision detection use case in a generic way 👍 

If you’re happy with that, I can go ahead and implement it that way.

Side note about the naming. In my view, `setApartBy` feels a bit more natural than `separated`. Feels more appropriate, because separation kinda implies there’s _something_ separating the two bits – or that they were once a whole and something has separated them. Happy to go with whatever you prefer though 👍

ianmackenzie:
My naming choice for `separation`/`separatedBy` was probably driven at least partially by the concept of a separating line/separating axis (https://en.wikipedia.org/wiki/Hyperplane_separation_theorem). I also think that 'separated' is just a generally more common word/phrase than 'set apart', so I think that's still my preference.

One thing to consider: is it clear what `overlappingBy 0` and `separatedBy 0` mean?

||overlapping boxes|touching boxes|separated boxes|
|---|---|---|---|
|`overlappingBy 0`|true since they are overlapping by at least zero, or false because the overlap is not zero?|true because the overlap is zero|true because there is zero overlap, or false because 'overlapping by at least zero' means at least touching?|
|`separatedBy 0`|true because there is zero separation, or false because 'separated by at least zero' means at most touching?|true because the separation is zero|true because they are separated by at least zero, or false because the separation is not zero?|

More generally: is it clear that `overlappingBy tolerance` means 'overlapping by at least the given tolerance'?

One idea: add an extra [`Order`](http://package.elm-lang.org/packages/elm-lang/core/5.1.1/Basics#Order) argument to `overlappingBy` and `separatedBy`! So the signatures would be

```elm
overlappingBy : Order -> Float -> BoundingBox2d -> BoundingBox2d -> Bool
separatedBy : Order -> Float -> BoundingBox2d -> BoundingBox2d -> Bool
```

and you'd call them as

```elm
-- see if two boxes are overlapping by more than 1e-3 units
-- (will return false if they are not overlapping at all)
if BoundingBox2d.overlappingBy GT 1e-3 firstBox secondBox then ...

-- see if two boxes are overlapping by at most 1e-3 units
-- (will return true if they are not overlapping at all)
if BoundingBox2d.overlappingBy LT 1e-3 firstBox secondBox then ...

-- see if two boxes have some non-zero overlap
-- (Tomek's original use case)
if BoundingBox2d.overlappingBy GT 0 firstBox secondBox then ...

-- see if two boxes are separated by at least 1e-3 units
-- (will return false if they are actually overlapping)
if BoundingBox2d.separatedBy GT 1e-3 firstBox secondBox then ...

-- see if two boxes are separated by less than 1e-3 units
-- (will return true if they are actually overlapping)
if BoundingBox2d.separatedBy LT 1e-3 firstBox secondBox then ...

-- see if two boxes have some non-zero separation
if BoundingBox2d.separatedBy GT 0 firstBox secondBox then ...

-- two equivalent ways to see if two boxes are exactly touching
-- (kinda weird but shouldn't be a common use case)
if BoundingBox2d.overlappingBy EQ 0 firstBox secondBox then ...
if BoundingBox2d.separatedBy EQ 0 firstBox secondBox then ...
```

(Note that `GT`, `LT` and `EQ` are all exposed as part of the `Basics` module so require no separate import.) There are no `GTE` or `LTE` values, but a hypothetical

```elm
BoundingBox2d.overlappingBy GTE 0 firstBox secondBox
```

would be equivalent to

```elm
BoundingBox2d.overlaps firstBox secondBox
```

or even

```elm
not (BoundingBox2d.separatedBy GT 0 firstBox secondBox)
```

As a side note, I think `overlaps` should probably be renamed to `intersects` - more symmetric with `intersection` and I think less likely to be misinterpreted as meaning 'actual finite overlap only'. `overlaps` can be kept for now as a deprecated alias until 3.0.

Phew - thoughts?

mpizenberg:
Having `intersection` and `intersects` with coherent behavior seems a good choice.

Regarding the following code example, either I'm a bit confused or you made a mistake. `overlappingBy EQ 0` just means that they are not strictly overlapping right? not that they are touching since this value is always >= 0
Same with `separatedBy`? You would have to check that both `EQ 0` to check if they are touching.

```elm
-- two equivalent ways to see if two boxes are exactly touching
-- (kinda weird but shouldn't be a common use case)
if BoundingBox2d.overlappingBy EQ 0 firstBox secondBox then ...
if BoundingBox2d.separatedBy EQ 0 firstBox secondBox then ...
```

Also, one could try to use `overlappingBy LT 0` which, in my understanding, will always return false.

ianmackenzie:
No, I think `overlappingBy EQ 0` would indeed mean touching - I think the semantics of `overlappingBy` (and similar for `separatedBy`) would be as if:

  - there was an `overlapDistance` function that returned a `Maybe Float`, which would be `Nothing` if the boxes didn't touch, `Just 0` if they just touched and `Just somePositiveValue` if they overlapped
  - `Maybe` values were `comparable`, where `Nothing` was less than any `Just` and `Just` values were compared based on their contents

Then `overlappingBy` could be implemented as

```elm
overlappingBy order amount firstBox secondBox =
    compare (overlapDistance firstBox secondBox) (Just amount) == order
```

so

  - `overlappingBy EQ 0` would only be true if the `overlapDistance` was equal to `Just 0`, i.e. the boxes just touch (since separated boxes would have an `overlapDistance` of `Nothing`)
  - `overlappingBy GT 1e-3` would be true if the `overlapDistance` was greater than `Just 1e-3`
  - `overlappingBy LT 1e-3` would be true if the `overlapDistance` was less than `Just 1e-3`; since `Nothing < Just 1e-3`, this would be true for separated boxes
  - `overlappingBy LT 0` would only return true for separated boxes, since `Nothing < Just 0`

Does that makes sense? Are there some cases I've missed?

mpizenberg:
Thanks, that makes perfect sense now.

ianmackenzie:
One more small tweak, I think: `overlappingBy` -> `overlapsBy`. A bit shorter (`overlapsBy GT 0` is actually one less character than `strictlyOverlaps`!), a bit more consistent with `intersects`, and reads a bit better in pipelines such as

```elm
firstBox |> BoundingBox2d.overlapsBy GT 1e-3 secondBox
```

@tomekwi what do you think of this latest proposed API? I'd be happy to take at stab at an initial implementation in the pull request branch if you'd like.

tomekwi:
Thanks for the insightful discussion @ianmackenzie @mpizenberg. I really like the API that we’ve come up with, especially considering the freedom that it gives us in terms of implementation.

If you could have a stab at the implementation, that would be ace @ianmackenzie! I’ve been thinking of tackling this myself, but I don’t trust myself to not mess up those tolerances. Especially with edge cases like when the shortest distance between bounding boxes is one between two corners not lying on the same axis.

Happy to jump in myself though if you’d rather just nudge me in the right direction 🙂

tomekwi:
> that we’ve come up with

That **you** guys have come up with of course 😄 I’m still learning my ways around the semantics and API patterns 🙂

ianmackenzie:
Sounds good @tomekwi! I'll try to take a stab at it some time in the next few days.

> That **you** guys have come up with of course 😄

No API springs from the void! There's no way we could have got to the current design without all the previous proposed variations and accompanying discussions.

ianmackenzie:
OK, just added initial implementations of `overlapsBy` and `separatedBy` - please take a look and see if you can poke any holes in the logic! I will try to add some more fuzz tests soon and then add `BoundingBox3d` versions (maybe `Interval` too while I'm at it).

mpizenberg:
Hi just some code thoughts I had while reading your implementation and related to the recently introduced `Interval` type. Not everything is "typely" correct but you can still get the idea.

```elm
overlapsBy : Order -> Float -> BBox -> BBox -> Bool
overlapsBy order tolerance box1 box2 =
    if tolerance >= 0 then
        case overlapAmount box1 box2 of
            Just distance ->
                compare distance tolerance == order
            Nothing ->
                order == LT
    -- tolerance < 0
    else
        (order == GT) && (overlapAmount box1 box2 /= Nothing)


type BoundingBox2d
    = BoundingBox2d Interval Interval


overlapAmount : BBox -> BBox -> Maybe Float
overlapAmount (BBox x1 y1) (BBox x2 y2) =
    let
        ( xOverlap, yOverlap ) =
            ( Interval.overlapAmount x1 x2
            , Interval.overlapAmount y1 y2
            )
    in
    case ( xOverlap, yOverlap ) of
        ( Just dx, Just dy ) ->
            Just (min dx dy)
        _ ->
            Nothing
```

ianmackenzie:
@mpizenberg that might very well be better - I optimized the current implementations for partial application so that cases like 

```elm
overlappingBoxes =
    List.filter (BoundingBox2d.overlapsBy GT 1e-3 someBox) candidateBoxes
```

were fast (checking of `Order` type and tolerance value are done when choosing what kind of function to construct, and then actual function application should be very fast) but I'm not sure if that's actually a good optimization in most cases.

Fun historical note: previous implementations of OpenSolid (in C++ and Scala) were more heavily interval arithmetic focused, so in both of those cases the `Interval` class was actually the *first* thing I implemented. As a result, in those implementations the bounding box types were indeed represented via component intervals as in your code, but I think I prefer the 'flatter' version of just storing min/max values directly (if nothing else it should be a bit more efficient since it involves fewer object allocations and less indirection for common operations like `intersects`). At some point I will probably have to do a long round of performance testing and optimization and then I may very well try out some other implementation ideas like the one you suggest!

In the meantime, @mpizenberg and @tomekwi could you try out the current implementation and see if there are any cases where it produces incorrect results? @tomekwi if you have some time and could add some hardcoded test cases, that would be awesome - I added some fuzz tests which seem to work, but I think it would be worth having a few hardcoded test cases as well.

src/OpenSolid/BoundingBox2d.elm
```diff
@@ -342,27 +361,271 @@ contains point boundingBox =
...
+module so are available by default in any Elm program.)
+
+Overlap is defined as the _minimum_ distance one box would have to move so that
+it did not touch the other, and is always positive for any two overlapping
+boxes.
+
+Boxes that just touch are considered to have an overlap of zero, which is
+distinct from 'no overlap'. Boxes that do not touch or overlap at all are
+considered to have an overlap which is less than zero but not greater or less
+than any negative number.
```
tomekwi: This left me wondering what it meant that

>Boxes that do not touch or overlap at all are considered to have an overlap which is less than zero but **not greater or less than any negative number**.

I don’t understand why “is less than zero” wouldn’t be an exhaustive description.

tomekwi: Oh, just after posting I read that sentence once more yet and it struck me what it means!

Perhaps we could add an encouraging phrase at the end of that paragraph, saying:

> Boxes that do not touch or overlap at all are considered to have an overlap which is less than zero but not greater or less than any negative number. **If you’re wondering what this means, the examples below will hopefully clear things up.**

ianmackenzie: Yeah, something like what you suggests sounds great! I'd say go ahead and make whatever tweaks like that you think are necessary and I can do a final pass at the end.

tomekwi:
Hey @ianmackenzie, I’ve finally managed to find a bit of time to tackle the stuff you suggested.

* I made a very small change to the bits of the docs which were confusing to me,
* added some hardcoded test cases for `separatedBy` (because the other stuff seemed pretty well tested)
* and updated some test names to match the naming of our new utils.

Let me know if there’s anything that you think I should improve.

ianmackenzie:
I just expanded on the manual test cases a bit - thanks for those and for the doc and test fixes!

Having thought over it a bit more I think I'm now leaning back towards `overlappingBy` instead of `overlapsBy` - more consistent with `separatedBy`, and I think I've convinced myself that pipelines like

```elm
if firstBox |> BoundingBox2d.overlappingBy GT 1e-3 secondBox then ...
```

read OK as "if the first box is overlapping by greater than 1e-3 units with the second box then...". Does anyone else care one way or the other or am I just bikeshedding by myself? 😁

tomekwi:
Fine with whichever works for you @ianmackenzie 😀

ianmackenzie:
OK, just switched back to `overlappingBy` and added `BoundingBox3d.[overlappingBy, separatedBy]` (no sense having those just for `BoundingBox2d`!). I think this is now ready to merge - any final thoughts?

I suspect there might be a way to make the docs a bit simpler/more concise, but I don't think that necessarily needs to happen within this PR.

src/OpenSolid/BoundingBox2d.elm
```diff
@@ -430,14 +430,15 @@ alwaysFalse firstBox secondBox =
 
 
 {-| Check if one box overlaps another by less than, greater than or equal to a
-given amount. You could perform a tolerant collision check (one that only
-returns true only if the boxes overlap by at least some small finite amount,
-and ignores boxes that just barely touch each other) as
+given amount. For example, you could perform a tolerant collision check (one
+that only returns true only if the boxes overlap by at least some small finite
```
mpizenberg: Double "only" here

ianmackenzie: Ah thanks, good catch!

src/OpenSolid/BoundingBox2d.elm
```diff
@@ -530,14 +531,16 @@ overlappingBy order tolerance =
 
 
 {-| Check if one box is separated from another by less than, greater than or
-equal to a given amount. If you were performing clash detection between some
-objects, you could use `separatedBy` on those objects' bounding boxes as a quick
-check to see if the objects had a gap of at least 1 cm between them:
+equal to a given amount. For example, if you were performing clash detection
```
mpizenberg: I think it would read better at present tense: "if you are performing ..., you can use ...". Or at least, it does not feel natural for me.

mpizenberg:
Just read the recent short commits (but not the one that trully matters with overlapping implementation sorry). These look good to me. I don't have much "strength" left to check the big one sorry. How confident are you on this port to 3D? ^^

ianmackenzie:
I'm pretty confident about the port to 3D - it was mostly very simple modifications. If you do want to look at one bit, I think the only function that changed substantially was `squaredSeparationAmount` (there are a greater number of possible cases in 3D, so an exhaustive `if` got a bit unwieldy).

mpizenberg:
Looks good to me

src/OpenSolid/BoundingBox2d.elm
ianmackenzie: Just pushed a slightly modified version ("For example, to perform clash detection between some objects, you could..."). Hopefully that reads a bit better!

ianmackenzie:
OK, I think this is ready to merge - thanks @tomekwi and @mpizenberg for all your work on this! I'll try to publish this as a minor release soon but I'd first like to finish the `EllipticalArc2d` module that I'm working on, since that's needed for [folkertdev/one-true-path-experiment](https://github.com/folkertdev/one-true-path-experiment).

tomekwi:
Thanks so much @ianmackenzie, it’s you who did the heavy lifting here man!

Will there be a PR for `EllipticalArc2d` that I could subscribe to to know when you’re done?

ianmackenzie:
@tomekwi hmm, no, I've just been working on `EllipticalArc2d` in `master`...I'll certainly let you know when it's done though! Do you find it a good practice in general to work in branches and submit pull requests to yourself? It also wouldn't be all that hard to push out a minor release earlier if that would be helpful.

tomekwi:
Thanks @ianmackenzie, no urgency at all. 🙂 It's just a matter of getting
some kind of reminder when things are ready to be cleaned up in my project.

As for pull requests for your own stuff, I think in a project managed by a
single person it doesn't matter much either way. 😇 I personally see value
in developing new features in branches, at least when the thing I'm
building takes up more than one commit. This gives me flexibility in
releasing master wherever it makes sense (not blocked by in-progress
stuff), makes it possible to abandon a thing if I hit a dead end and also
open a pull request as needed (for example to get feedback from people).

Issue #26, opened 2017-10-12
Add Interval sin/cos


ianmackenzie:
This is the last remaining piece of work from #1. I have an old implementation in Scala starting [here](https://github.com/ianmackenzie/opensolid-geometry-scala/blob/53af12c87e45494a8fa379d2e48098242261acd4/src/main/scala/org/opensolid/core/Interval.scala#L495) that may be useful.

```elm
module Interval exposing (...sin, cos, ...)

sin : Interval -> Interval
cos : Interval -> Interval
```

ianmackenzie:
Implemented in #30 - thanks @danfishgold!

PR #30, opened 2017-10-14
Add functions to Interval

danfishgold:
This handles #26 but not like the Scala implementation.
I documented the functions and the most error prone part is probably `Interval.cosIncludesMax`.

I hope this helps 😄

### DIFF
--- a/src/OpenSolid/Interval.elm
+++ b/src/OpenSolid/Interval.elm
@@ -285,3 +285,13 @@ isContainedIn : Interval -> Interval -> Bool
 isContainedIn firstInterval secondInterval =
     (minValue secondInterval >= minValue firstInterval)
         && (maxValue secondInterval <= maxValue firstInterval)
+
+
+isSingleton : Interval -> Bool
+isSingleton (Interval { minValue, maxValue }) =
+    minValue == maxValue
+
+
+shift : Float -> Interval -> Interval
+shift delta (Interval { minValue, maxValue }) =
+    Interval { minValue = minValue + delta, maxValue = maxValue + delta }
--- a/src/OpenSolid/Interval.elm
+++ b/src/OpenSolid/Interval.elm
@@ -295,3 +295,81 @@ isSingleton (Interval { minValue, maxValue }) =
 shift : Float -> Interval -> Interval
 shift delta (Interval { minValue, maxValue }) =
     Interval { minValue = minValue + delta, maxValue = maxValue + delta }
+
+
+sin : Interval -> Interval
+sin ((Interval { minValue, maxValue }) as interval) =
+    if isSingleton interval then
+        singleton (Basics.sin minValue)
+    else
+        let
+            ( includesMin, includesMax ) =
+                sinIncludesMinMax interval
+
+            newMin =
+                if includesMin then
+                    -1
+                else
+                    min (Basics.sin minValue) (Basics.sin maxValue)
+
+            newMax =
+                if includesMax then
+                    1
+                else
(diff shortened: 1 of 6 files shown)


ianmackenzie:
Thanks for the PR! I've added some tests for `sin` and `cos` (which required adding a few features and doing some reorganization that I've been meaning to do anyways), and everything seems to work!

I've made a few small modifications (mostly to make things a bit more consistent with other code in the package) but no real functional changes.

I've hidden `shiftBy` for now since I think that makes more sense to expose as part of a set of general interval arithmetic functions like `scaleBy`, `sum`, `difference`, `product`, and `quotient`. I'm a bit on the fence whether the `Interval` module *should* support general interval arithmetic, since using a chain of interval operations makes it a bit too easy to generate too-large bounds on some value and in many cases a more specialized approach would be better. (For example, you could determine bounds on a rotated bounding box by doing a bunch of interval arithmetic on the X, Y and Z intervals of the original box, but you'd get a much larger result than if you just rotated each vertex of the box and then found the hull of the resulting points.) If we do add general arithmetic operations to the `Interval` module, then I think it might make sense to split that out into its own small `opensolid/interval` package to make it more easily reusable in non-geometric contexts (see https://github.com/opensolid/meta/issues/2).

Do the changes all look good? If so, then I think this is ready to merge! (By the way, out of curiosity, how did you find out about this package/how are you using it?)

danfishgold:
Everything looks great! I wanted to suggest a few things but you covered all of them.

I found the package by searching for Elm projects on GitHub that have open issues with the hacktoberfest tag, so thanks for doing that. I haven't used the package yet, but hopefully some day I'll get into WebGL stuff :)

ianmackenzie:
Great - merging! I'm working on some WebGL packages right now, hopefully they're ready by the time you get into WebGL stuff =)

@mpizenberg thanks for the suggestion to tag issues with Hacktoberfest!

src/OpenSolid/Interval.elm
```diff
@@ -285,3 +324,129 @@ isContainedIn : Interval -> Interval -> Bool
...
+    , interval |> cosIncludesMax
+    )
+
+
+{-| The maximum of cos(x) is x = 2 pi * k for every integer k.
+If `minValue` and `maxValue` are in different branches
+(meaning diffrent values of k), then the interval must pass through
+2 pi * k, which means the interval must include the maximum value.
+-}
+cosIncludesMax : Interval -> Bool
```
mpizenberg: There is also the edge case when `min = 2 * k * pi` and `max < 2 * (k+1) * pi`. Both return the same `floor`. That probably could be checked by `floor ( minValue / (2*pi )) == ceiling ( minValue / (2*pi) )` ? Not sure if this is important or if it can happen.

ianmackenzie: That's a good point, but I think in this case it's fine - if `cosIncludesMax` or one of its relatives returns `False`, then the function is evaluated at both endpoints anyways. So it should only really be important that `cosIncludesMax` catches cases that are strictly *inside* the given interval.

ianmackenzie:
Oh - @danfishgold I totally forgot to ask you to add yourself to the AUTHORS file! Feel free to open a new pull request doing that, or just let me know how you'd like to be listed and I can add you in. If you'd prefer not to be listed, that's fine too, but don't be shy =)

danfishgold:
I guess `Dan Fishgold <d.fishgold@gmail.com>` fits the existing format 😄

Issue #11, opened 2016-10-24
Add Sphere3d type and module


ianmackenzie:
`type Sphere3d = Sphere3d { centerPoint : Point3d, radius : Float }`

This would be very similar to the `Circle2d` type and module, so that could be used as an example/template. A couple of interesting `Sphere3d`-specific additions:

```elm
projectOnto : Plane3d -> Sphere3d -> Circle3d
projectInto : SketchPlane3d -> Sphere3d -> Circle2d
```

daphee:
What exactly would projectOnto do? I guess something like

```
projectOnto : Plane3d -> Sphere3d -> Circle3d
projectOnto plane sphere =
        Circle3d.with 
             { centerPoint = Point3d.projectOnto plane (centerPoint sphere)
             , axialDirection = Plane3d.normalDirection plane
             , radius = ???
             }
```

'm not sure about how the radius would transform.

ianmackenzie:
Yes, exactly, and `radius = Sphere3d.radius sphere` since this is just a simple orthographic projection, not something like a perspective projection which would distort things. Do you think the docs should make that more clear?

daphee:
I think "orthographically" is better than perpendicularly, mainly because the Wikipedia page on "orthographical projection" has a nice immediately visible picture which can make clear what the projection does. 😃 

I'll need to write some more tests and will create a PR in the next few days. 

One more question: are you using elm-format? if yes, do you have some customized settings? The reason I ask is that I have my editor set to automatically apply elm-format and now in the files I edited the bodies of all `let` ... `in` ... expressions are shifted to the right by 4 spaces.

daphee:
>  I'm not sure it extends to projection onto axes, though - would you keep "perpendicular" for that?

Hmm... Not sure on that one. On the one hand it is nice to have everything consistent. On the other hand "perpendicular" is the more common / ""easier"" word. 

> In the future I'll try to be more careful to use the stable/standard version of elm-format, but in the meantime if you could use 0.7.0-exp for OpenSolid stuff that would be great. Thanks, and sorry for the annoyance!

Cool, I'll do that. I'm not doing anything else with Elm at the moment so it shouldn't be any problem to switch to the 0.7.0-exp version.

ianmackenzie:
Implemented in #32 - thanks @daphee!

Issue #27, opened 2017-10-12
Convert documentation examples to tests


ianmackenzie:
I started converting a bunch of the examples shown in the documentation to tests; for example see commits 8025fe78ba15395cd07a0a410c0240c8e2e11282 and be3dbb11f58d2d79fe702aa40bc033163f25537c. It would be great if someone could help converting over more of them! As nice as it would be, I don't think it's feasible to use something like [stoeffel/elm-verify-examples](https://github.com/stoeffel/elm-verify-examples) since most tests in this package require using custom comparison functions that include a small numerical tolerance for roundoff.

The general procedure would be to find examples in a particular module, convert them to tests in the corresponding tests file, then run the tests locally using [`elm-test`](https://github.com/elm-community/elm-test) to verify that they pass.

daphee:
Hi, I'd like to help you with the tests. 

I'm new to the repo but after a few glances into various files it seems like there's lots of examples, which is great, but converting them and keeping them in sync manually seems pretty tedious. I played with elm-verify-examples a little but couldn't find a way to use a custom comparison function as well.

I opened an [issue](https://github.com/stoeffel/elm-verify-examples/issues/39) asking to change to syntax of elm-verify-examples to include a way to specify a custom comparison function.

ianmackenzie:
Thanks, the issue looks great! I've gone and added a few comments. I agree that figuring out a way to get `elm-verify-examples` to work in this use case is a much better approach.

daphee:
I have some good news!

I couldn't stop thinking about how I would extend the functionality of elm-verify-examples to cover this use case. Obviously I liked my idea of introducing a line prefixed by `|` containing the function to be used between the assertion and expectation part.

An example from Arc2d:

```
    Arc2d.endPoint exampleArc
    | Expect.point2d
    --> Point2d.fromCoordinates ( 1, 3 )
```

So I went on and implemented that. I am not sure whether it really is a good idea to put this there because it definitely obfuscates the examples a little bit and makes them harder to read.

That's what I wanted your opinion on and that's also the reason why I haven't submitted a PR to elm-verify-exampley yet.

I have adjusted `Arc2d` to use the new syntax on my [branch](https://github.com/daphee/geometry/tree/verify-examples) so you can see it in action!

Maybe we should explore using RegExes in seperate configuration files but I fear that those will get quite complicated as e.g. they have to check not only for Point3d/Arc2d but also whether Maybe is used. Still this way the documentation wouldn't get cluttered.

[...]

Issue #18, opened 2017-03-09
Potential breaking changes for 2.0


ianmackenzie:
This is a meta-issue for potential breaking changes to make for a 2.0 version. Feel free to post suggestions in the comments and I'll edit this issue text.

## Module naming

I used prefixed module names like `OpenSolid.Point2d` to avoid potential naming conflicts with modules from other packages, but based on elm-lang/elm-compiler#1625 it seems that this might be considered "silly module renaming". Should the prefixes be removed? I think they should remain in a few cases like `OpenSolid.Geometry.Types` and `OpenSolid.Geometry.Decode`, but otherwise module names could be switched to just plain `Point2d`, `Triangle2d` etc.

This would be more consistent with other Elm packages - `elm-community/elm-test` uses top-level `Expect`, `Fuzz` and `Test` modules, `mdgriffith/style-elements` uses top-level `Style` and `Element` modules, `terezka/elm-plot` uses a top-level `Plot` module, etc.

Additionally, if the non-prefixed names _don't_ work out and run into conflicts with other packages, then these conflicts will provide additional data points for elm-lang/elm-compiler#1625.

## Renames

I used things like `radialDistanceFrom` to disambiguate from plain `distanceFrom` (which takes another point as argument) while keeping the nice 'read as a phrase' quality. However, I've come to think these names are a little too 'cute'/obscure and it would probably be better to use more 'boring' names, even though `Point3d.distanceFromAxis axis` doesn't read quite as nicely as `Point3d.radialDistanceFrom axis`.

  - `Point3d.radialDistanceFrom` -> `Point3d.distanceFromAxis`
  - `Point3d.squaredRadialDistanceFrom` -> `Point3d.squaredDistanceFromAxis`
  - `Point3d.projectRadiallyOnto` -> `Point3d.projectOntoAxis`

## Removals

  - `Direction#d.scaleBy`: use `Vector#d.in_` instead (sounds a bit weird to 'scale a direction')
  - Perhaps move primitive types (points, vectors, directions, axes, planes, frames, bounding boxes) to their own `opensolid/primitives` package? Could make them more appealing for people who don't want to depend on a large package or (shudder) 'framework'...
[...]

Issue #12, opened 2016-10-24
Add Torus3d type and module


ianmackenzie:
`type Torus3d = Torus3d { axis : Axis3d, majorRadius : Float, minorRadius : Float }`

eskimoblood:
It would be easier to have on base function that can create a mesh from a formula, in the end the only different between a torus a sphere or a supershape is the [formula](https://github.com/eskimoblood/elm-parametric-surface/blob/master/src/Surface/Formulars.elm) that creates the points from the two angles and a bunch of predefine parameters. I've started to [implement](https://github.com/eskimoblood/elm-parametric-surface) this in Elm, but at the moment its just a fun project. I would really like to port it over to opensolid and make a reusable library out of it.

ianmackenzie:
For the purposes of meshing, you are correct, but that is not the only possible use of a 3D geometry library! There are several functions that I could see adding to this module that would be specific to toruses:

  - Checking if a point is within a torus
  - Finding the intersection (if any) between a line segment and a torus
  - Determining the closest point on a torus to a given point

It's entirely possible that some meshing code could be shared between toruses and spheres, but to me that seems more like an implementation detail than a design decision. Thoughts?

eskimoblood:
For me it looks like there are solids with its settings and then there is stuff you can do with solids like rendering or checking for intersection. So you can define a torus type with its parameters and put it into the `createMeshFromSolid` or the `isPointInSolid` function.

ianmackenzie:
Sorry for the delayed response! It's been a busy week at work...

At some point I think it is likely there will be a `Body3d` type, but I think it still makes sense to have the specific types like `Sphere3d`, `Torus3d` etc. I envision something like

```elm
module OpenSolid.Body3d exposing (Body3d, contains)

import OpenSolid.Torus3d as Torus3d exposing (Torus3d)
import OpenSolid.Sphere3d as Sphere3d exposing (Sphere3d)

type Body3d
    = Torus Torus3d
    | Sphere Sphere3d

contains : Point3d -> Body3d -> Bool
contains point body =
    case body of
        Torus torus ->
            Torus3d.contains point torus

        Sphere sphere ->
           Sphere3d.contains point sphere
```

[...]

PR #32, opened 2017-10-29
Sphere3d

daphee:
This adds a `Sphere3d` module, like discussed in #11.
I closely followed the functionality of Circle2d and Circle3d when implementing it. 

I just wrote tests for three methods so far but at least `throughPoints`, the most complicated method, is covered.

It is also the first module that is completely verifiable using stock `elm-verify-examples`.

`elm-verify-examples` does have functionality to declare intermediate values and include them in the generated tests (e.g. see `localFrame`) but because `exampleSphere` is used in every example and because I didn't want to copy and paste its definition into every comment I added it to the file `tests/Docs/DocTestStatic/Sphere3d.elm` and imported it.

The examples also require a few other imports. I bundled them into an additional non-documentation comment near the top. `elm-format` would remove one of the four spaces until I added another `-` after `{-`. I think most of the imports could be added to the examples without confusing the reader of the generated documentation. The only import which would seem out of place is the one importing `exampleSphere` from `tests/Docs/DocTestStatic/Sphere3d.elm`. Maybe we should explicitly define `exampleSphere` in every example instead?


.gitignore
```diff
@@ -9,6 +9,9 @@ elm.js
 documentation.json
 docs.json
 
+# elm-verify-examples generated tests
+tests/Doc/OpenSolid
```
ianmackenzie: Very minor point but couldn't this just be tests/Doc? I don't anticipate changing the OpenSolid name but tests/Doc seems a little more direct...just 'ignore the doc tests'!

daphee: The reason I didn't to tests/Doc is that there is tests/Doc/DocTestStatic. So we would need to either get rid of that or move it up into tests

ianmackenzie: Ah, I see, there's some non-generated code inside the Doc directory

ianmackenzie:
This looks great! I'm going through now and making a few changes here and there, but nothing major yet.

One thing I'm a bit on the fence about is the `roundTo` functions - they fit in pretty well with the rest of the API and they're certainly useful for doc tests, but I'm worried that they could encourage bad coding practices. Specifically, I've often heard people hit floating point equality issues and say "oh, let's just round everything to 3 digits and THEN check for equality" - that works if you have 1.234001 and 1.233999 (both round to 1.234), but not if you have 1.234499 and 1.234501 (the former rounds to 1.234 and the latter rounds to 1.235 even though the difference betwen them is only 0.000001). Might be enough to highlight that in the docs, though - what do you think?

.gitignore
ianmackenzie: Oops, didn't refresh the page (so didn't see your reply) until I posted mine, sorry =)

daphee:
I agree, I am not very happy about the equality checking in general as well.

- I think roundTo could be useful in other contexts as well, e.g. when whether a sphere contains a point there is currently no way to specify a tolerance. This way a call to roundTO could be chained before the call to compare.

- This doesn't invalidate your point but, in your example do you think using the `truncate` function would be more sensible? Maybe a `truncateTo` function should be added as well?

- `Sphere3d` is missing a `equalWithin` function. Maybe it's best if the examples would use equalWithin as well.

- I preferred the "official" `Expect` module functions over the reimplementations in `OpenSolid.Geometry.Expect`. For example one benefit of using

ianmackenzie:
> I think `roundTo` could be useful in other contexts as well, e.g. when whether a sphere contains a point there is currently no way to specify a tolerance. This way a call to `roundTo` could be chained before the call to compare.

What exactly would the code look like for this? If you mean something like `Point3d.roundTo 2 >> Sphere3d.contains` instead of just `Sphere3d.contains`, then some points would get rounded away from the sphere and some would get rounded towards - you'd end up with "check if the point is contained within a voxel approximation of the sphere" instead of "check if the point is contained within the sphere".  I think this is a good example of why `roundTo` might *not* be a good thing to support/encourage.

> `Sphere3d` is missing a `equalWithin` function. Maybe it's best if the examples would use `equalWithin` as well.

The reason that I didn't add `equalWithin` for anything more complex than points is that the semantics get a bit ambiguous. For example, the simplest approach for `Sphere3d.equalWithin` would be

```elm
equalWithin : Float -> Sphere3d -> Sphere3d -> Bool
equalWithin tolerance firstSphere secondSphere =
    Point3d.equalWithin tolerance (centerPoint firstSphere) (centerPoint secondSphere)
        && Scalar.equalWithin tolerance (radius firstSphere) (radius secondSphere)
```

but if two spheres had their center points separated by `tolerance` *and* their radii differed by `tolerance`, then you'd have some points on one sphere that were `2 * tolerance` away from the other sphere (which seems wrong if we've asked the two spheres to be equal within `tolerance`). A better definition might be enforcing that no point on one sphere is more than `tolerance` away from the other sphere (by checking whether the distance between center points *plus* the absolute difference between radii is less than `tolerance`). But the fact that the semantics of `Sphere3d.equalWithin` are less obvious than `Point3d.equalWithin` makes me uneasy.

ianmackenzie:
> I preferred the "official" Expect module functions over the reimplementations in OpenSolid.Geometry.Expect. For example one benefit of using Expect.within over OpenSolid.Geometry.Expect.within is that in the case of an error the formatting is prettier.

The official `Expect.within` didn't exist when I created the `OpenSolid.Geometry.Expect` module - funny you should bring that up, I actually just noticed the new official version recently and switched the code over to using it, but foolishly did it as part of #21 which has not yet been merged. So feel free to use the official version since everything else will be using it soon. Are there any other duplicate `Expect` functions you know of?

daphee:
> What exactly would the code look like for this? If you mean something like Point3d.roundTo 2 >> Sphere3d.contains instead of just Sphere3d.contains, then some points would get rounded away from the sphere and some would get rounded towards 

Hmm, your right. So you should only use `roundTo` if for example your intent is to get the final value of a lengthy calculation to include it in a written up report or something like that, right?

So far I used `roundTo` twice: Once in the example for `circumference` and once in the example for `rotateAround`. I think the latter is a clear use case for `truncateAt` because the need to use any rounding/truncating arises from inaccurate floating point math (e.g 0.0e-12 where the expected result would actually be 0) and it would be quite intuitive to just cut the number at the 11th digit. 

In the example for `circumference` it should be okay to use `roundTo` but again truncate probably reflects the intent more clearly.

Are you okay with changing those two to a new `truncateTo` function?

> you'd end up with "check if the point is contained within a voxel approximation of the sphere" instead of "check if the point is contained within the sphere". I think this is a good example of why roundTo might not be a good thing to support/encourage.

Isn't that sort of a limitation of the floating point datatype though?

Have you thought about using a differnt floating point type internally? Maybe something that is similar to Python's `decimal` module (which I have never used). Every time a number needs to be returned to the user the normal `Float` could be used but internally decimals would be saved with higher/arbitrary precision. I don't have enough experience with Elm to know if this would be technically feasible and don't have enough experience with this library to know if it would even make sense but I just wanted to bring it up.

ianmackenzie:
>  So you should only use roundTo if for example your intent is to get the final value of a lengthy calculation to include it in a written up report or something like that, right?

Exactly - it's really only appropriate for display purposes, I can't think of any cases where it's useful as part of logic (and there are lots of places where it *seems* like a good approach but isn't). The more I think about it the more it seems that `roundTo` should be left to dedicated string formatting libraries such as `krisajenkins/formatting` which actually [already has a `roundTo` function](http://package.elm-lang.org/packages/krisajenkins/formatting/4.2.0/Formatting#roundTo). I'd prefer to look at cases where `roundTo` seems like a good solution and try to provide some better-behaved functions that solve the higher-level problem in a safer/less error-prone way. (I don't think `truncateTo` is any better, really.)

> Have you thought about using a different floating point type internally? Maybe something that is similar to Python's decimal module (which I have never used).

There are lots of approaches around to dealing with precision/roundoff, like [MPFR](http://www.mpfr.org/) for variable-precision floating point values or Jonathan Shewchuk's [fast robust predicates](https://www.cs.cmu.edu/~quake/robust.html) for exact operations on standard floating-point values. However, trying to implement something like that in Elm and use it in OpenSolid would incur a massive performance hit (I'd estimate it would be hundreds or even thousands of times slower than just using `Float` operations, which can get JIT-compiled down into single machine instructions). So I think sticking with `Float` as the underlying data type and being careful to provide operations that are numerically well-behaved is the way to go. It's certainly possible to get excellent results using just `Float` values - there's a *lot* of precision available in a 53-bit mantissa - and I think this package has a role to play in guiding people towards well-behaved numeric/geometric code in the same way that the Elm Architecture guides people towards well-structured UI code.

I think this may mean splitting out the discussion of doc tests from the `Sphere3d` implementation itself - it might even make sense to try implementing something similar to `elm-verify-examples` that's designed specifically for how OpenSolid examples are written.

Side note: I'm travelling for the next few days so responses might be delayed (and it generally takes me a while to work through pull requests anyways - check out the amount of back-and-forth on some of the other ones!). From my interpretation, for Hacktoberfest it's only important that the pull request is *made* in October, not *merged*, right? I don't want to be the difference between a Hacktoberfest t-shirt and no Hacktoberfest t-shirt =)

daphee:
> I'd prefer to look at cases where roundTo seems like a good solution and try to provide some better-behaved functions that solve the higher-level problem in a safer/less error-prone way. (I don't think truncateTo is any better, really.)

Yeah that makes sense. So e.g. providing a `containsWithTolerance : Tolerance -> Point3d -> Sphere3d -> Bool` function, right?

> I think this may mean splitting out the discussion of doc tests from the Sphere3d implementation itself - it might even make sense to try implementing something similar to elm-verify-examples that's designed specifically for how OpenSolid examples are written.

I think it would be great to write actual unit tests in the same nice and concise fashion the examples are written. Is that part of what you mean with the above quote?

> and it generally takes me a while to work through pull requests anyways

This might sound like a dumb question but: is it even worth for you to handle a big PR then? I guess it is but how could I make my code as easy to review as possible?

Is there anything I can do to help on this PR right now?

> From my interpretation, for Hacktoberfest it's only important that the pull request is made in October, not merged, right?

Yeah, thanks :grin: I *do* hope I can help out here even after Hacktoberfest. I find the project quite interesting, enjoyed writing this PR and tbh there aren't lots of good Elm projects in general and even less that are as welcoming to newcomers.

What would be a good next issue to work on?

ianmackenzie:
Hey, sorry for the delay, back from travelling now =)

> So e.g. providing a `containsWithTolerance : Tolerance -> Point3d -> Sphere3d -> Bool` function, right?

Exactly, assuming `Tolerance` is an alias for `Float` (I also think I'd call it `containsWithin`). I think this is something I'd prefer not to add right away, though - I think even a `containsWithin` function might not be the best approach for a lot of problems, so I'd prefer to wait until it actually comes up once or twice and then look at the use cases, especially since right now it's pretty easy to use something like

```elm
point |> Point3d.distanceFrom (Sphere3d.centerPoint sphere) <= Sphere3d.radius sphere + tolerance
```

which is a bit wordy but I think very clear and explicit.

> This might sound like a dumb question but: is it even worth for you to handle a big PR then?

PRs are certainly still useful! For one thing, I do often struggle with RSI in my hands from too much typing, so I'd rather spend two hours reading/reviewing code than one hour writing it. Apart from mundane advantages like that, though, it's really valuable to just see how other people approach geometric problems in different ways, get new ideas for features/improvements, and just generally discuss design/implementation issues with other people so I'm not coding in a bubble. I'm also fairly new to trying to manage external contributions, so I'm extra cautious right now - I suspect that over time I'll come up with better and more efficient workflows that allow me to merge PRs more quickly and efficiently, but I can't do that without going through a bunch of PRs first =)

> > I think this may mean splitting out the discussion of doc tests from the Sphere3d implementation itself - it might even make sense to try implementing something similar to elm-verify-examples that's designed specifically for how OpenSolid examples are written.
>
> I think it would be great to write actual unit tests in the same nice and concise fashion the examples are written. Is that part of what you mean with the above quote?

I think what I'd like is if this PR could be amended to just use 'standard' doc examples and tests for now, and continue the discussion of automatic doc tests separately. This will allow the `Sphere3d` functionality itself to be merged in sooner rather than later (I'd actually love to have it in for some work I'm doing on a separate 3D graphics package), and we can continue the doc tests discussion at our leisure - there are enough different possibilities there that I don't think I'm comfortable committing to one right away. I'll add some more comments to #27 and perhaps we can continue the discussion there for now?

daphee:
> I think what I'd like is if this PR could be amended to just use 'standard' doc examples and tests for now

I think the fact that every example other than the *one* that needs `Sphere3d.roundTwo` are verifiable using the "off-the-shelf" version of `elm-verify-example` is a cool benefit to get "for free". 

I get that it is better to think thorougly how this problem can be solved long-term but on the other hand adjust just that one example (e.g. maybe so that it doesn't even need the roundTo function) and including a elm-verify-example check into Travis wouldn't be much effort (or is it? I never used Travis before :D ) . This doesn't mean that all examples should be converted to be compatible just not "looking a gift horse in the mouth". 

So I would:

- remove Sphere3d.roundTo (keep the Scalar one)
- change the one example that uses it
- include elm-verify-examples into the checks Travis performs (in the short term and for new modules only)

What do you think?

ianmackenzie:
> adjust just that one example (e.g. maybe so that it doesn't even need the `roundTo` function) and including a elm-verify-example check into Travis

Sure, that sounds good - I confess I haven't looked at the actual `Sphere3d` module yet so I didn't know how much the examples were different. I'm totally fine with adding an `elm-verify-examples` step to the Travis config (happy to do that myself, should be easy).

>  - remove `Sphere3d.roundTo` (keep the `Scalar` one)

Is there a strong reason to keep `Scalar.roundTo`? I think it's a bit dangerous in the same way `Point3d.roundTo` is - my preference would be to leave all of the `roundTo` functions out for now, go as far as we can with 'vanialla' `elm-verify-examples` and then have a separate discussion about modifying `elm-verify-examples`/figuring out different ways to write tests/writing a custom doc-to-test converter etc.

tests/Circle3d.elm
```diff
@@ -22,6 +25,49 @@ jsonRoundTrips =
...
+    Test.describe "throughPoints"
+        [ Test.fuzz3 Fuzz.point3d
+            Fuzz.point3d
+            Fuzz.point3d
+            "All given points lie on the circle constructed using `throughPoints`"
+            (\p1 p2 p3 ->
+                let
+                    -- if the first three points are collinear it is not possible to construct a circle
+                    -- that passes through them
+                    isValidInput =
```
ianmackenzie: I think this could be done more succinctly and robustly by a triangle area check - if the area of the triangle formed by the three points is zero, then the points are collinear. Probably also check for area being below a tolerance (say 1e-6) instead of exactly zero, to filter out really degenerate cases where there will be bad roundoff.

tests/Circle3d.elm
```diff
@@ -22,6 +25,49 @@ jsonRoundTrips =
...
+                                (\axis ->
+                                    Point3d.squaredDistanceFromAxis axis p3 /= 0
+                                )
+                            |> Maybe.withDefault False
+
+                    circle =
+                        Circle3d.throughPoints ( p1, p2, p3 )
+
+                    liesOnCircle point circle =
+                        Point3d.squaredDistanceFrom point (Circle3d.centerPoint circle)
```
ianmackenzie: I think using `Point3d.squaredDistanceFrom` here is a premature optimization - `Point3d.distanceFrom` is much more clear (especially when tolerances are involved) and it's not like we're calling it on a million different points.

tests/Sphere3d.elm
```diff
@@ -0,0 +1,191 @@
...
+                                Point3d.squaredDistanceFromAxis axis p3 /= 0
+                            )
+
+                -- it is not possible to construct a sphere from four points that are all in the same plane
+                notInTheSamePlane =
+                    Plane3d.throughPoints ( p1, p2, p3 )
+                        |> Maybe.map (flip Point3d.signedDistanceFrom p4)
+                        |> Maybe.map ((/=) 0)
+
+                isValidInput =
```
ianmackenzie: Similarly here I think you should be able to do a tetrahedron volume check - there's no `Tetrahedron3d` type but you can still calculate the volume with something like

```elm
v2 = Vector3d.from p1 p2
v3 = Vector3d.from p1 p3
v4 = Vector3d.from p1 p4
volume = abs (Vector3d.dotProduct v4 (Vector3d.crossProduct v2 v3))
```

src/OpenSolid/Sphere3d.elm
```diff
@@ -0,0 +1,500 @@
...
+
+    Sphere3d.around Axis3d.z point
+    --> Sphere3d.with
+    -->     { centerPoint =
+    -->         Point3d.fromCoordinates ( 0, 0, 2 )
+    -->     , radius = 3
+    -->     }
+
+-}
+around : Axis3d -> Point3d -> Sphere3d
```
ianmackenzie: I'm a bit hesitant about including `around` since for any given axis and point there are an infinite number of spheres that that are 'around' the given axis (have their center point on the axis) and through the given point (as opposed to circles, where there's only one solution).

I might be tempted to have

```elm
fromGreatCircle : Circle3d -> Sphere3d
fromGreatCircle circle =
    with
        { centerPoint = Circle3d.centerPoint circle
        , radius = Circle3d.radius circle
        }
```

instead (which you could chain with `Circle3d.around` to get the current `Sphere3d.around` behavior), or just leave it out for now and see what use cases come up that need something like this.

src/OpenSolid/Sphere3d.elm
```diff
@@ -0,0 +1,500 @@
...
+            if y /= 0 then
+                Just <|
+                    with
+                        { centerPoint = Point3d.along normalAxis d
+                        , radius = sqrt (r ^ 2 + d ^ 2)
+                        }
+            else
+                Nothing
+        )
+        (Circle3d.throughPoints ( p1, p2, p3 ))
```
ianmackenzie: Minor point but I think piping would make this a bit more readable:

```elm
Circle3d.throughPoints ( p1, p2, p3 )
    |> Maybe.andThen
        (\circle ->
            ...
        )
```

src/OpenSolid/Sphere3d.elm
```diff
@@ -0,0 +1,500 @@
...
+                r =
+                    Circle3d.radius circle
+
+                x =
+                    Point3d.distanceFromAxis normalAxis p4
+
+                y =
+                    Point3d.distanceAlong normalAxis p4
+
+                d =
```
ianmackenzie: I think I'd move this down into a separate `let` block inside the first branch of the `if y /= 0` expression, so that it's clear that you only divide if the denominator is non-zero

src/OpenSolid/Sphere3d.elm
```diff
@@ -0,0 +1,500 @@
...
+
+    Sphere3d.circumference exampleSphere
+    -> 18.8496
+
+-}
+circumference : Sphere3d -> Float
+circumference sphere =
+    2 * pi * radius sphere
+
+
```
ianmackenzie: Can you add `surfaceArea` as well? =)

src/OpenSolid/Sphere3d.elm
```diff
@@ -0,0 +1,500 @@
...
+        { centerPoint = Point3d.scaleAbout point scale (centerPoint sphere)
+        , radius = abs (scale * radius sphere)
+        }
+
+
+{-| Rotate a sphere around a given axis by a given angle (in radians).
+
+    exampleSphere
+        |> Sphere3d.rotateAround Axis3d.y (degrees 90)
+    -> Sphere3d.with
```
ianmackenzie: Can you update this to use the standard multiline `-->` format?

src/OpenSolid/Sphere3d.elm
```diff
@@ -0,0 +1,500 @@
...
+
+    Sphere3d.contains
+        (Point3d.fromCoordinates (4.00001, 2, 1))
+        exampleSphere
+    --> False
+
+-}
+contains : Point3d -> Sphere3d -> Bool
+contains point sphere =
+    Point3d.squaredDistanceFrom point (centerPoint sphere) <= radius sphere ^ 2
```
ianmackenzie: I'm a bit nervous about using `^` since it (unsurprisingly) looks like it compiles down to a `Math.pow` call, which seems generally slower than a simple multiplication, especially on older browsers (see https://stackoverflow.com/questions/26593302/whats-the-fastest-way-to-square-a-number-in-javascript). That's why I use a temporary `r` variable in similar places - perhaps a premature optimization, but I think the result is still quite readable...

ianmackenzie:
Just added a bunch of comments to the `Sphere3d` module itself - see what you think! (In a few places it probably would have been easier for everyone for me to just make the changes myself, but once I'd started adding comments through the web interface it was simpler to just keep doing that...)

src/OpenSolid/Sphere3d.elm
daphee: This was intentional so that it wil not be verified by `elm-verify-example`. I can try to find an example that leads to round numbers.

tests/Circle3d.elm
daphee: Done. 

Out of interest I set `isValidInput = True` and it seems the fuzzer pretty much never creates three collinear  points. Maybe we should investigate the fuzzer and how to get better "coverage" a little bit in the future. An easy solution would be to just always do both: fuzz based testing and testing using hard coded test cases.

src/OpenSolid/Sphere3d.elm
daphee: 👍

src/OpenSolid/Sphere3d.elm
daphee: I removed it for now.

daphee:
I implemented the changes you proposed. I'm not sure what to do with the examples that currently don't pass when tested with `elm-verify-examples` and which I have "disabled" by using `->` instead of `-->`.

ianmackenzie:
Awesome, thanks!

One alternate approach for the documentation is something I've been thinking about for a while - trying to use more visual instead of textual examples. I've written this up as a new issue in #33.

In the interim, brainstorming here...how hard do you think it would be to post-process the doc test failures to see which ones are *actual* failures and which are just roundoff issues? A bit messy but might be doable...would be pretty easy to add some sort of post-processing script to the Travis config. Basically I'm thinking of something that would look for "expected ..., got ..." in the test output and then do something like extract all the numerical values from the two strings and compare them pairwise with a tolerance.

daphee:
> extract all the numerical values from the two strings and compare them pairwise with a tolerance.

I think this could actually be done with vanilla elm-verify-examples. You can import additional functions. This also means that you can override Expect.equal. So we could override Expect.equal to compare arbitrary objects by using the string comparison you described on the String representation from `toString`. 

This is a very hacky solution but tbh I still think it is the best short term one. Figuring out how to do proper comparisons with tolerance is too difficult to try to solve it as quick as possible.

ianmackenzie:
Ha, that's very clever! If you're willing to prototype that solution and see if it works, that would be great.

daphee:
I think it works =) Only problem is that I can't "shadow" the already imported Expect module with my custom one.

```
This usage of variable `Expect.equal` is ambiguous.

289|                     Expect.equal
                         ^^^^^^^^^^^^
Maybe you want one of the following?

    Expect.equal
    Doc.DocTestStatic.Expect.equal
```

If I manually remove the `import Expect` module it works as expected.

ianmackenzie:
Awesome! I don't mind a little bit of ugliness/trickiness if it's really just confined to just the Travis config, and I think this should otherwise let us have our cake and eat it too (examples that read nicely with no extra cruft or special associated comments, but that can be automatically tested). This setup means that you generally wouldn't run the doc tests on local checkouts, but I think that's totally fine - running on every push via Travis should be more than enough.

What would you think of renaming your function `equalWithinTolerance` or `equalByToString` or something explicit like that and then have your `sed` command replace `Expect.equal` with `Expect.equalWithinTolerance` instead of removing the `import Expect` line? Might be slightly cleaner/less tricky...

Also, apologies for the occasional spurious test failures...I've logged that as an issue in #34.

daphee:
Yeah, I agree that it makes the purpose of the function more clear. I do think that it is more tricky though because more can go wrong when we are replacing all `Expect.equals` instead of just a single line that's always the same. I still changed it like you proposed.

Cool! Can this get merged then? =)

ianmackenzie:
I think it's getting close! I'd like to do one final pass through the code, but then I think it might indeed be ready to merge.

You asked earlier what would be a good next issue to work on - what sort of things are you most interested in? There's certainly lots more geometric functionality that would be useful in this package (toruses #12, cylinders #10, intersection functions #4 and #5, interval-valued functions #3) but it would also be great to just extend the doc test coverage to more existing modules (which will probably require some tweaks here and there). I also have a handful of other related packages for working with SVG, WebGL and generic triangular meshes but those are all either much smaller or under rapid/early development where collaboration is s bit trickier.

daphee:
Thanks for remembering! I think I'll work on extending the test coverage.

- I'll try to verify all examples as soon as this is merged. 
- A good start would probably be to increase `Sphere3d` test coverage before merging. So far only `with`, `throughPoints` and `boundingBox` are tested. 
- I think ideally even small methods like `Sphere3d.circumference` would have some tests, right? Those tests would be very similar to the example tests in scope but it's probably still a good idea to have some proper tests "set in stone" and not as a "side-effect" of the documentation.

daphee:
Sorry for the double comment but I just remembered something I noticed even in the beginning: why is there no e.g. `Circle3d.circle3d : Point3d -> Direction3d -> Float -> Circle3d` constructor? I think in some cases it would be easier to be able to construct objects that way instead of using records.

tests/Sphere3d.elm
```diff
@@ -202,6 +202,70 @@ properties =
         )
 
 
+scaleAbout : Test
+scaleAbout =
```
daphee: I am not sure whether those tests actually provide any value.

ianmackenzie:
Test coverage: sounds great! And yes, over time the general test coverage should certainly be increased. So far I've only been testing more complex stuff where I'm less sure about the implementation.

To the extent that I had a plan for increasing test coverage it was to do it in parallel with performance enhancements. Currently most functions are implemented in a pretty direct, readable, intuitive way - for example I don't think there's *too* much room for bugs in stuff like

```elm
signedDistanceAlong : Axis3d -> Point3d -> Float
signedDistanceAlong axis point =
    Vector3d.from (Axis3d.originPoint axis) point
        |> Vector3d.componentIn (Axis3d.direction axis)
```

but it does allocate a temporary `Vector3d` value and make a handful of function calls, both of which can be slow (especially allocation). If it turns out that function is a common bottleneck, it would make sense to rewrite the implementation to manually inline everything to just extract scalar coordinates/components and then directly compute the signed distance - should be possible to do without *any* allocations or function calls! But then there's a lot more room for typos (using `x`/`y`/`y` instead of `x`/`y`/`z`, for example), so I figured any time I did an optimization like that I'd add a test for the function to make sure I hadn't messed anything up. No harm in adding those tests *before* doing the performance optimizations, though!

> Sorry for the double comment but I just remembered something I noticed even in the beginning: why is there no e.g. Circle3d.circle3d : Point3d -> Direction3d -> Float -> Circle3d constructor? I think in some cases it would be easier to be able to construct objects that way instead of using records.

Hehehe - this has definitely come up before, see the discussion on #18 (search for instances of lower-case `point3d` on that page to find most of the relevant comments). In general I agree that functions like

```elm
Circle3d.circle3d : Point3d -> Direction3d -> Float -> Circle3d
Point3d.point3d : Float -> Float -> Float -> Point3d
```

would be more convenient in many cases (especially when quickly creating hardcoded values, or doing JSON decoding). On the other hand:

  - `Point3d.fromCoordinates`, `Circle3d.with` etc. let you follow a consistent pattern of "all functions are prefixed with the module name" (it seems to me that `point3d` and `circle3d` functions would be meant to be imported unqualified)
  - I think it's nicer and more understandable when modifying code/looking at diffs - switching from `Point3d.fromCoordinates` to `Point3d.fromPolarCoordinates` is pretty understandable and easy to follow, while switching from `point3d` to `Point3d.fromPolarCoordinates` is a bit more jarring
  - Most importantly, I do like the explicitness and clarity of having "named arguments" by using records, and I'm willing to sacrifice some conciseness to get it - perhaps because I'm most interested in ultimately using this package to do engineering/design work, where (as opposed to games etc.) correctness is of absolute paramount importance

I've toyed around with the idea of having a separate module containing functions like `point3d`, `circle3d` etc. - call it `OpenSolid.Geometry.Curried` or `OpenSolid.Geometry.Basics` or `OpenSolid.Geometry.Concise` or something, similar to how [Evan recommends putting infix operators in an `Infix` module](http://package.elm-lang.org/help/design-guidelines#avoid-infix-operators). But I'd prefer there to be exactly one canonical way to (for example) construct a `Circle3d` from its center point, axial direction and radius, so that code written by one person is as readable as possible to someone else who's also familiar with the package.

tests/Sphere3d.elm
```diff
@@ -202,6 +202,70 @@ properties =
...
+rotateAround : Test
+rotateAround =
+    fuzz3 axis3d
+        scalar
+        sphere3d
+        "A sphere gets rotated correctly"
+        (\axis angle sphere ->
+            sphere
+                |> Sphere3d.rotateAround axis angle
+                |> Expect.all
```
daphee: Do you prefer `Expect.all` (this test) or `Expect.equal (Sphere3d.with)` (scaleAbout)?

ianmackenzie:
Ooh, maybe `OpenSolid.Geometry.Unqualified`...if we do ultimately go down that route that's not a bad name. (I'd still prefer *not* to, though.)

```elm
import OpenSolid.Geometry.Unqualified exposing (..)
```

daphee:
> increasing test coverage it was to do it in parallel with performance enhancements

I think this maybe answers one of the comments I added to the commits I just pushed. The test would no longer be a straight copy of the actual function but rather a slower, more readable version. 

> search for instances of lower-case point3d on that page to find most of the relevant comments

Lol, I choose `Point3d.point3d` only because I couldn't come up with a better name. I don't actually mind prefixing functions with their package, I think it makes code a lot easier to read. What I do like though are curryable functions. I understand your reasoning though and think it is in line with the Elm philosophy: Keep things simple. Similar to that I think the tests could be a lot more concise if the type system were a little bit more advanced.

tests/Sphere3d.elm
ianmackenzie: I guess between the two I slightly prefer the `Expect.equal` version since it seems a bit simpler/more readable, although I think I'd use `Expect.sphere3d` instead to allow a tiny bit of tolerance - small differences in implementation could result in rounding differences between the test and implementation. The `Expect.all` version seems more flexible, though, for checking various properties of the result. So maybe `Expect.sphere3d` etc. where that's all you need, and `Expect.all` if you're doing something more complex?

In general, I think it's generally considered best practice to try *not* to have the test mirror the implementation - for example, the `Point3d.rotateAround` tests check for stuff like "the original point and rotated point should be the same distance from and along the rotation axis", not "the coordinates of the resulting point are equal to this rotation matrix times the coordinates of the original point", and you could add more elaborate stuff like "if you construct the direction from the axis to the original point and from the axis to the rotated point, the angle between the two directions should be equal to the absolute value of the given rotation angle" (at least for rotation angles less than 180 degrees). In this case I don't see a way to do anything like that, so I think what you've done is fine, but it's something to consider for other tests in the future.

ianmackenzie:
For what it's worth I have tried to make functions curried wherever I think that makes sense. For example, I don't think it makes a lot of sense to create a bunch of points at different Z heights by doing something like

```elm
List.map (point3d 1 2) [ 3, 4, 5 ]
--> [ point3d 1 2 3, point3d 1 2 4, point3d 1 2 5 ]
```

(what if you wanted points with different X or Y values instead?). I *do* think it makes sense to supply only the start point of a line segment to do stuff like get radial lines from one point to a bunch of other points, so there's the curried `LineSegment3d.from` in addition to the single-argument `LineSegment3d.fromEndpoints`:

```elm
List.map (LineSegment3d.from p0) [ p1, p2, p3 ]
--> [ LineSegment3d.from p0 p1
--> , LineSegment3d.from p0 p2
--> , LineSegment3d.from p0 p3
--> ]
```

It's less obvious to me how you would curry something like `Circle3d.with` - off the top of my head it seems most likely you'd want the center point argument *last* so you could use `List.map` to create a bunch of circles with the same radius/orientation at different points. I'm certainly willing to consider adding curried versions of functions where the semantics are clear and there's a reasonable use case.

ianmackenzie:
Hey @daphee just a quick note to say that I haven't forgotten about this - events have conspired to keep me from doing much lately, so I haven't been able to do a proper review. But I will as soon as I can!

daphee:
Thanks for the update. No problem at all, I intend to improve the tests for `translateBy` and `mirrorAcross` but I was too busy as well.

ianmackenzie:
Finally got some time to take a proper look at this - although first I had to track down a Windows incompatibility in `elm-verify-examples` (see stoeffel/elm-verify-examples#42) that was preventing me from running the doc tests on my laptop. (By the way, any chance you could check out my fork of `elm-verify-examples` from that PR and see if it works for you? I don't have a Linux machine to test on, and the Travis build is timing out - I don't know if that's a result of my change or just something random.)

I've moved the `DocTestStatic` stuff into a new `src/OpenSolid/Geometry/Examples` directory. I tried to think of some good organizational principles for stuff like this, and I think what I'd like to try out is "importable code goes within `src/OpenSolid` wherever possible" - I think that makes imports easier to read.

I also made a couple small tweaks to the `Sphere3d` module and modified a few of the tests to use tolerant comparisons - I think in general I default to tolerant comparisons to make the tests robust to changes in the implementation.

By the way, I really like some of the tests you've added for `rotateAround` etc.! Shall I leave this PR open so you can make the improvements to the `translateBy` and `mirrorAcross` tests that you  mentioned? Otherwise I think this is ready to merge...

daphee:
> By the way, any chance you could check out my fork of elm-verify-examples from that PR and see if it works for you?

I checked out your branch, ran `bash run-tests.sh` and got `16` passes and `0` fails. So I'd guess it's a random problem with the Travis build. Or is there something else I'd need to check?

> I've moved the DocTestStatic stuff into a new src/OpenSolid/Geometry/Examples directory.

Cool, I was pretty unsure about where to put those files anyway.

I improved the tests for `mirrorAcross` and `translateBy` a little as well. If you are okay with those changes it really should be ready to merge.

ianmackenzie:
Updated tests look great! And thanks for checking the `elm-verify-examples` branch - I think you're right that that means it's probably some unrelated Travis thing.

Merging - thanks so much for all your work on this! And if you're still interested in expanding test coverage with future PRs, that would be awesome =)

PR #39, opened 2018-01-19
Adds a convex hull implementation

gampleman:
I took a quick stab at this today - this is a Monotone chain algorithm.

I’m not sure this is the right place or API name.


ianmackenzie:
This looks great, thanks! I've made a few tweaks and added a visual test - see what you think. (To get the visual test working, you'll need to check out `opensolid/svg` into a directory beside your `opensolid/geometry` checkout so the relative paths in `media/elm-package.json` work.)

For the function name, I think `Polygon2d.convexHull` probably makes the most sense (it's constructing a polygon which _is_ a convex hull, not constructing a polygon from an existing convex hull). Alternatively, I'd be tempted to move it to `Point2d.convexHull`, but in general I prefer to have simpler modules like `Point2d` not have too many dependencies on more complex modules like `Polygon2d`. Thoughts?

ianmackenzie:
Oops, forgot to update the test...but I was planning on switching to a fuzz test anyways, so I'll go ahead and do that now.

ianmackenzie:
Oh, also: can you add yourself to the AUTHORS file? =)

ianmackenzie:
By the way, is there a description somewhere of the algorithm you used? It would be nice to have a link to it in a comment in `fromConvexHull` (not a doc comment, just something for anyone in the future looking at the code).

gampleman:
OK, I've made the suggested changes.

Do you think having a function that would indicate whether a point is inside a polygon would make sense? Seems like that would be useful as well?

ianmackenzie:
A point-inside-polygon test would certainly be useful, but I think that's separate issue with its own tradeoffs - I've added it as #40.

I'm still a bit undecided on `Polygon2d.convexHull` vs `Point2d.convexHull`, but otherwise I think this is ready to merge! I like the unambiguity/readability of `Point2d.convexHull`, but the circular dependency is a bit messy and would get in the way of potentially splitting the 'primitive' types into their own package (opensolid/meta#1). I can decide that later, though - shall I go ahead and merge this?

gampleman:
👍

PR #53, opened 2018-07-08
Fix typo in README

infertux:


ianmackenzie:
Hi @infertux! Good eye, and apologies if this tripped you up. The actual story is a bit more complex though - I've been preparing for Elm 0.19 in the `master` branch, and in Elm 0.19 `elm install` _is_ the correct command. The Elm 0.18 version is in the `elm-0.18` branch, and that's what I've been publishing from, so the README at http://package.elm-lang.org/packages/ianmackenzie/elm-geometry/latest is correct.

However, it's definitely confusing that just going to the main GitHub page of `elm-geometry` shows you a README for a future version of Elm! In hindsight I probably should have kept 0.18 code in `master` and done 0.19 preparations in an `elm-0.19` branch. I could probably do some branch renaming, but for now I've just changed the default branch on GitHub to `elm-0.18` so you should see a valid README by default. How does that sound?

infertux:
Gotcha. This makes perfect sense. Thanks for the detailed explanations. I wasn't aware that Elm 0.19 changed the command - I'm still using 0.18.

Thank you for making this package and elm-3d-camera BTW. They are saving me a lot of time :)

Issue #60, opened 2018-07-31
Add BoundingBox#d.expandBy


ianmackenzie:
`BoundingBox#d.expandBy : Float -> BoundingBox#d -> Maybe BoundingBox#d`

Use negative value to shrink, return `Nothing` if shrunken past zero width - can default to the original box or `BoundingBox#d.singleton (BoundingBox#d.centroid box)` if desired

MrL1605:
I'm picking this issue up and I have a doubt relating to this issue.
If boundingBox is represented by : min (3,2) max (8,6) 
And it is expanded by 3
Then resulting boundingBox is represented by : (0,-1) (11,9)

Let me know if I'm incorrect.

ianmackenzie:
That's correct!

ianmackenzie:
Added in #82. Thanks @MrL1605!

PR #82, opened 2018-10-13
Added expand by function to BoundingBox#d

MrL1605:
Added function expand by to BoundingBox2d and BoundingBox3d
Code is reformatted using correct version of `elm-format`.
Related Issue: #60  

Pull request checklist:
  - Is the pull request from a dedicated branch in your fork instead of `master`?   **Yes**
  - Have you added yourself to the AUTHORS file? **Yes** 
  - Is "Allow edits from maintainers" enabled? **Yes**
    `elm-geometry`? **Yes**
  - Is code wrapped to 80 columns? **Yes**
  - **Have you added tests for new functionality?** **No**
  - Have you added documentation for new functionality? **Yes**

### DIFF
--- a/src/BoundingBox2d.elm
+++ b/src/BoundingBox2d.elm
@@ -12,7 +12,7 @@ module BoundingBox2d exposing
     , fromExtrema, singleton, from, hull, intersection, aggregate, containingPoints
     , extrema, minX, maxX, minY, maxY, dimensions, midX, midY, centerPoint, centroid
     , contains, isContainedIn, intersects, overlappingBy, separatedBy
-    , scaleAbout, translateBy, translateIn
+    , scaleAbout, translateBy, translateIn, expandBy
     )
 
 {-| A `BoundingBox2d` is a rectangular box in 2D defined by its minimum and
@@ -52,7 +52,7 @@ box of an object than the object itself, such as:
 
 # Transformations
 
-@docs scaleAbout, translateBy, translateIn
+@docs scaleAbout, translateBy, translateIn, expandBy
 
 -}
 
@@ -922,3 +922,34 @@ is equivalent to
 translateIn : Direction2d -> Float -> BoundingBox2d -> BoundingBox2d
 translateIn direction distance boundingBox =
     translateBy (Vector2d.withLength distance direction) boundingBox
+
+
+{-| Expand the bounding box in all the directions by given distance;
+
+
+    expandBy_ : Float
+    expandBy_ =
+        3
+
+
+    --> BoundingBox2d.expandBy expandBy_ exampleBox
+    -->     { minX = 0
+    -->     , maxX = 11
+    -->     , minY = -1
+    -->     , maxY = 9
+    -->     }
+
+-}
+expandBy : Float -> BoundingBox2d -> Maybe BoundingBox2d
+expandBy by boundingBox_ =
+    let
+        centroidPt_ =
+            centroid boundingBox_
+    in
+    Just <|
+        fromExtrema
+            { minX = minX boundingBox_ - by
+            , minY = minY boundingBox_ - by
+            , maxX = maxX boundingBox_ + by
+            , maxY = maxY boundingBox_ + by
+            }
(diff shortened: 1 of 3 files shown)


src/BoundingBox2d.elm
```diff
@@ -922,3 +922,52 @@ is equivalent to
...
+
+
+{-| Expand the bounding box in all the directions by given distance;
+
+    expandBy_ : Float
+    expandBy_ =
+        3
+
+
+    --> BoundingBox2d.expandBy expandBy_ exampleBox
```
ianmackenzie: I don't think the `expandBy_` variable is necessary, I think `BoundingBox2d.expandBy 3 exampleBox` would be shorter and arguably more readable...

ianmackenzie:
First of all, as I mentioned in your other PR, apologies for the late reply! I don't think using the half diagonal length is actually the correct logic here...for example, if you had a bounding box of width 8 and height 6 then the half diagonal length would be 5. But shrinking by 5 (expanding by -5) would give you an invalid bounding box! The real limit should be 'half of the width or half of the height, whichever is less' (in this case, 3).

That said, I'm not sure you really have to compute that limit directly - I think it might be simpler to calculate `newMinX = minX box - by`, `newMaxX = maxX box + by` etc., then have a final `if` expression that checks that `newMinX <= newMaxX` etc. If the box is shrunken too much, then one of those checks will fail and you can return `Nothing`.

The other option would be to just do nothing if given a negative value. Then `expandBy` wouldn't have to return a `Maybe`, and there could be a separate `offsetBy` that _did_ accept negative offsets (and did return a `Maybe`). That is, the current function would be renamed to `offsetBy` and there would be a new `expandBy` function that would return a plain `BoundingBox#d`. Thoughts?

MrL1605:
No worries for delayed response. 

I'm sorry for the diagonal logic, didn't verify if it was correct (that's a big silly mistake). I'll update according to the half the width or height logic.

I think your second option is clean as `expandBy` should only handle expansion, and `offsetBy` could do both, so that developers don't need to use `Maybe` even when they are sure they are expanding it.

I'll do this in a today.

src/BoundingBox2d.elm
MrL1605: Done

ianmackenzie:
Hey @MrL1605, sorry for the delay, just getting back into the swing of things after having moved. Just tweaked the documentation (I kind of like writing the docs myself, to make sure everything has a consistent style).

I think I'd like to try switching `expandBy` to use the absolute value of the argument instead of ignoring negative values (so if you passed -3 the box would expand by 3). This is pretty consistent to how some other `elm-geometry` functions work.

Finally, I think there are still a few issues with the code itself:

  - I would use `( width, height ) = dimensions boundingBox_` instead of `dimensions_ = dimensions boundingBox_`; then you can just use `smallerDimension = min width height` instead of an `if` and a bunch of `Tuple` functions
  - More importantly, the code looks like it's checking against `smallerDimension` when it should really be checking against `smallerDimension / 2` (and I think the `if` and `else if` could be replaced by just a single `if by > -smallerDimension / 2`)
  - It's a bit messy to call `offsetBy` in `expandBy` when you know it will succeed; might be good to have an `unsafeOffsetBy` function (not exposed publicly) that you could call from both `offsetBy` and `expandBy`

MrL1605:
I'll add mentioned changes.

MrL1605:
I'm really sorry if this was more effort on your side than required.
I'll make sure to comment on any issue before starting to work on it. 
A sincere thank you for the efforts, got to learn a lot of things.

ianmackenzie:
No apology necessary, I'm still trying to figure out exactly how things should work! Glad it's been helpful for you =) On that note, you may have noticed these from my last commits but a couple other small notes you may find useful:

  - You can use `-x` instead of `-1 * x`
  - Elm has a built in `abs` function

ianmackenzie:
By the way, check out https://package.elm-lang.org/packages/elm/core/latest/Basics to see what other useful functions are available by default in Elm!

Issue #64, opened 2018-08-14
Add centroid functions


ianmackenzie:
- [x] `Point#d.centroid : List Point#d -> Maybe Point#d`
- [ ] `Polygon2d.centroid : Polygon2d -> Maybe Point2d` (weight triangle centroids by signed area)
- [x] `Polyline#d.centroid : Polyline#d -> Maybe Point2d` (weight edges by length)

davcamer:
@ianmackenzie Are you actively working on this? I am thinking of doing it as part of Hacktoberfest 🎃

ianmackenzie:
@davcamer I am not - thanks for checking first! If you're willing to have a go, that would be great =)

I've actually just started a pretty huge refactoring (in a branch) to update `elm-geometry` to use my recently-released `elm-units` package for all coordinate etc. values, but I'm starting with low-level stuff like points and vectors. It will probably be a while before I get to polylines and polygons, so there shouldn't be any issue merging your changes into my branch before that happens. Even if I do get to polylines/polygons before you finish, any merge conflicts should be very easy to resolve, so please don't feel you need to rush.

davcamer:
👍 

I was looking at some of the Hacktoberfest issues in `elm-units` as well, but I should stick to one at a time. They are both nice libraries.

I will start with Polygon centroids.

ianmackenzie:
A couple implementation comments/suggestions off the top of my head:

  - Should return `Nothing` if the polyline/polygon has no points
  - `Polyline#d.centroid` will be a bit simpler than `Polygon2d.centroid` (but do whichever you'd prefer!)
  - For polylines, the simplest implementation would basically be to add up weighted line segment centroid coordinates and then divide by total weight at the end, but this would encounter roundoff issues if the coordinates were large (the polyline was far from the origin). A better approach would be:
    - Find the rough centroid by getting the centroid of the polyline's bounding box
    - Add up the vectors from that approximate centroid to each line segment's centroid, weighted by line segment length
    - Divide the vector sum by overall weight, then add to the initial approximate centroid
  - Similar for polygon centroid - for that you'll end up forming a bunch of triangles (one for each polygon edge), and it would be better to build those triangles from an approximate bounding box-based centroid instead of from the global origin

davcamer:
Thank you for the suggestions. As soon as I started looking at polygon centroid implementations I started to think polylines might have been the better choice. Particularly the advice about avoiding overly large values I don't think I would have come across without your input.

[...]

PR #83, opened 2018-10-14
Polyline{2,3}d centroid function

davcamer:
The implementation uses the approach suggested by Ian on the original
issue:
https://github.com/ianmackenzie/elm-geometry/issues/64

Here is his description:
For polylines, the simplest implementation would basically be to add
up weighted line segment centroid coordinates and then divide by total
weight at the end, but this would encounter roundoff issues if the
coordinates were large (the polyline was far from the origin). A better
approach would be:

* Find the rough centroid by getting the centroid of the polyline's
  bounding box
* Add up the vectors from that approximate centroid to each line segment's
  centroid, weighted by line segment length
* Divide the vector sum by overall weight, then add to the initial
  approximate centroid

Pull request checklist:

  - Is the pull request from a dedicated branch in your fork instead of
    `master`? (Preferred, but not mandatory.)
  - Have you added yourself to the AUTHORS file? Not mandatory if you'd prefer
    to stay anonymous, but don't be shy!
  - Is "Allow edits from maintainers" enabled? (There should be a checkbox for
    this when you open the pull request.)
    `elm-geometry`? (If running `elm-format` doesn't change any existing code,
    then the answer is probably 'yes'.)
  - Is code (mostly) wrapped to 80 columns? (It's OK if type annotations and
    occasional things like string literals are longer.)
  - BONUS POINTS: Have you added tests for new functionality?
  - EXTRA BONUS POINTS: Have you added documentation for new functionality?
    (Don't worry too much about documentation text - I like that to have a
    consistent style, which means writing most of it myself - but if you can
    come up with some code examples to use in documentation that would be
    fantastic.)

Don't worry too much if you're not sure if one of the above is true, or if you
know it's not true but aren't sure how to fix it - just go ahead and open the
pull request and we'll sort it out. The intent is for this list to be helpful,
not intimidating!


davcamer:
I noticed there is a `Polygon2d.triangulate` method. Are you suggesting calculating triangles from the bounding box center because it will be more performant than triangulating via `TriangularMesh`?

ianmackenzie:
Yeah, `Polygon2d.triangulate` is very complex internally (necessarily so!), so won't be nearly as efficient as the approach using signed areas. Check out [Polygon2d/Monotone.elm](https://github.com/ianmackenzie/elm-geometry/blob/master/src/Polygon2d/Monotone.elm) if you want to see what happens when you take a complex geometric algorithm designed for imperative code with arrays and pointers and translate it into Elm =)

Just looking through the updated code now...

ianmackenzie:
Oh btw, thanks for catching the `examplePolyline` 2D/3D typo in `Polyline3d.elm` =)

ianmackenzie:
One super minor nit (I'd just make the change myself, but I'm on the bus with my phone): can you change `\_ ->` to `\() ->` in the various `Test.test` tests? The former accepts any argument, while the latter only accepts the unit value `()`. This means it conveys a bit more information when read, and adds a bit of type safety - if `Test.test` changes in the future to pass a useful argument to the lambda, then you'll get a compile error instead of the lambda accepting the new argument and then silently ignoring it.

ianmackenzie:
I think everything else looks good! Once the tiny lambda change is in, do you think this is ready to merge?

davcamer:
Interesting point about the params for the test function. A co-worker and I had been talking about this a few days earlier and hadn't come up with a reason to prefer one over the other, but I like your point about future changes. Cheers!

I think it is ready to merge now. I'll do the Polygon centroids on another branch.

ianmackenzie:
Great, thanks! The type safety with `\() ->` is definitely nice, but if you ask me, the "extra information when read" aspect is even more important. If I see `\() ->` when reading a code base then I immediately know that it's being used for lazy evaluation of some sort, while `\_ ->` could also mean that there's an ignored argument and I'd have to look closer to see. (When I was reviewing your changes I actually did end up checking the `elm-test` documentation quickly to verify that it wasn't passing anything useful to the actual test function.)

Issue #51, opened 2018-06-20
Add 'midpoint' functions to curve types


ianmackenzie:
Example signature:

```elm
CubicSpline2d.midpoint : ArcLengthParameterized -> Point2d
```

MrL1605:
I would like to work on this issue. 
Let me know if there are any other inputs.

MrL1605:
I would like some more help in this.

I tried to search for some resources as I didn't know how to calculate midpoints of curves myself, but couldn't find much.

If anyone has any resources for parameterized curves I can refer to, it would be great.

MrL1605:
Following link is useful for Cubic Spline 
http://blog.sklambert.com/finding-the-control-points-of-a-bezier-curve/#finding-the-control-points

MrL1605:
Hi @ianmackenzie, 
The signature you suggested includes ArcLengthParameterized
```ELM
CubicSpline2d.midpoint : ArcLengthParameterized -> Point2d
```
In case of CubicSpline2d I only need the end points and control points. 
Check changes in [above PR](https://github.com/ianmackenzie/elm-geometry/pull/85/commits/4323ec87b79200f329c80eb06439ce2fa2fc1b08#diff-fdd7fab72dfb5d5f41d11e168148810eR981)

As creating ArcLengthParameterized requires one to call build, it would be optimal to use following signature instead.
```ELM
CubicSpline2d.midpoint : CubicSpline2d -> Point2d
```

PS: I'll add more midpoint functions to other curves as well

ianmackenzie:
Hey @MrL1605 I've replied in #85 - probably easiest to continue the discussion there.

ianmackenzie:
Spline curve midpoint functions added in #85 - now just need circular and elliptical arc versions.

PR #85, opened 2018-10-14
Midpoint functions for curves

MrL1605:
Added function midpoint to CubicSpline2d and CubicSpline3d
Code is reformatted using correct version of `elm-format`.

Related Issue: #51 

Pull request checklist:
  - Is the pull request from a dedicated branch in your fork instead of `master`?   **Yes**
  - Have you added yourself to the AUTHORS file? **Yes** 
  - Is "Allow edits from maintainers" enabled? **Yes**
    `elm-geometry`? **Yes**
  - Is code wrapped to 80 columns? **Yes**
  - **Have you added tests for new functionality?** **No**
  - Have you added documentation for new functionality? **Yes**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@ elm.js
 documentation.json
 docs.json
 
+# Idea directory created by Intellij Webstorm
+.idea/
+
 # elm-verify-examples generated tests
 tests/VerifyExamples
 
--- a/src/CubicSpline2d.elm
+++ b/src/CubicSpline2d.elm
@@ -975,7 +975,7 @@ arcLength parameterizedSpline =
 {-| Get the midpoint of the spline.
 
     CubicSpline2d.midpoint parameterizedSpline
-    --> Point2d.fromCoordinates  (3.999999999999992, 2.5)
+    --> Point2d.fromCoordinates (3.999999999999992, 2.5)
 
 -}
 midpoint : ArcLengthParameterized -> Point2d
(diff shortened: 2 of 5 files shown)


ianmackenzie:
Hi @MrL1605,

First of all, apologies for the delayed response - I'm in the middle of a move, so I've had limited access to my computer.

I think you may have made this a bit more complex than necessary - I think all that's actually required is something like

```elm
midpoint : ArcLengthParameterized -> Point3d
midpoint parameterized =
    case pointAlong parameterized (arcLength parameterized / 2) of
        Just point ->
            point

        Nothing ->
            -- Should never happen since half of total arc length will always
            -- be a valid distance along the curve, but let's default to something
            -- reasonable anyways
            startPoint (fromArcLengthParameterized parameterized)
```

It looks like what you're trying to do is use [De Casteljau's algorithm](https://en.wikipedia.org/wiki/De_Casteljau%27s_algorithm) (which is what `CubicSpline3d.pointOn` etc. use) to find the midpoint, but evaluating at a spline parameter value of 0.5 is *not* in general the same as finding the point physically/geometrically halfway along the curve (halfway by actual arc length along the curve). For that you *do* need an arc length parameterization. Does that make sense?

I apologize for not making this clearer in the issue description - I have not been nearly as careful to add detailed descriptions/explanations to issues that I haven't flagged as 'Hacktoberfest', 'good first issue', 'help wanted' etc., they're really been meant more as just brief reminders to myself.

MrL1605:
Understood. I'll update it.

MrL1605:
Thanks for the clarification. 
I guess there wasn't much after your example, but if there are any more changes required, please let me know.

ianmackenzie:
Apologies for the delay - looks good!

Issue #40, opened 2018-01-22
Add point-in-polygon test


ianmackenzie:
Probably best approach is to use winding number: http://geomalgorithms.com/a03-_inclusion.html

However, this is still linear time in the number of polygon edges. Is there a more efficient approach for large, complex polygons? One idea would be to construct a triangulation of the polygon, store the triangles in an R-tree and then use that to determine which (if any) of the triangles a given point is contained in. However, for (say) the triangulation of a polygon approximating a circle (with many thin triangles crossing the entire polygon) there may still be O(n) triangles whose bounding box overlaps with any given point.

gampleman:
AFAIK currently there is nothing that enforces the polygon vertices to be sorted, right? 

Won't this be a problem here? Because if we need to sort them, then we'll need to pay O(nlogn)...

ianmackenzie:
I should probably add a note to the docs...constructors like `Polygon2d.fromVertices` **do** expect the vertices to be given in order (either clockwise or counterclockwise is OK, but random order is not). And I don't see how it could be otherwise - from a list of vertices in random order there's no way to tell which polygon was intended. For example, four vertices in a square with one vertex in the middle of the square might represent

```
o---o
 \  |
  o |
 /  |
o---o
```

or they might represent

```

o---o
|   |
| o |
|/ \|
o   o
```

and there's no way to determine automatically which one the user wants. Or am I misinterpreting your comment?

gampleman:
That makes perfect sense. Perhaps a note in the docs would prevent others from asking such a stupid question :)

ianmackenzie:
Added in #96.

PR #96, opened 2019-04-02
Point in polygon test

gampleman:
Based on:

> Based on Hao, J.; Sun, J.; Chen, Y.; Cai, Q.; Tan, L. Optimal Reliable Point-in-Polygon Test and Differential Coding Boolean Operations on Polygons. Symmetry 2018, 10, 477.

The code comes pretty directly from the paper and as such is very highly optimized. Specifically it attempts to minimize the number of operations needed and used geometric probability to structure the code. As such, it isn't the prettiest, but is quite straightforward. I've left in comments that specify which case each branch refers to, but to decode what this means, one must refer to the paper:

Cases

### DIFF
--- a/tests/Tests/Polygon2d.elm
+++ b/tests/Tests/Polygon2d.elm
@@ -12,7 +12,7 @@ import Geometry.Expect as Expect
 import Geometry.Fuzz as Fuzz
 import LineSegment2d
 import Point2d
-import Polygon2d exposing (convexHull)
+import Polygon2d
 import Test exposing (Test)
 import Triangle2d
 import TriangularMesh
--- a/tests/Tests/Polygon2d.elm
+++ b/tests/Tests/Polygon2d.elm
@@ -1,5 +1,6 @@
 module Tests.Polygon2d exposing
-    ( convexHullContainsAllPoints
+    ( containsPointTest
+    , convexHullContainsAllPoints
     , convexHullIsConvex
     , triangulationHasCorrectArea
     , triangulationHasCorrectNumberOfTriangles
@@ -10,7 +11,8 @@ import Fuzz
 import Geometry.Expect as Expect
 import Geometry.Fuzz as Fuzz
 import LineSegment2d
-import Polygon2d
+import Point2d
+import Polygon2d exposing (Containment(..), convexHull)
 import Test exposing (Test)
 import Triangle2d
 import TriangularMesh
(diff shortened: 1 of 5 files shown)


ianmackenzie:
Very cool! I'll try to take a more detailed look this evening, but a couple questions to get started:

- Does this work for non-convex polygons? Polygons with holes?
- Is it actually important to distinguish between 'boundary' and inside/outside? My philosophy in general is that numerical roundoff makes it unsafe to check for things like "exactly on the boundary" (since that's equivalent to checking for floating point equality), so that's not something that should be encouraged/supported. My inclination would be to simply treat 'boundary' as 'contained' and then just return a `Bool`. Thoughts?

gampleman:
> Does this work for non-convex polygons? Polygons with holes?

Should do (according to the paper it handles everything including self-crossing and nested holes). I haven't tested non-convex, but the unit tests include holes.

Regarding boundaries: I think those become important for doing the boolean stuff that comes later in the paper (although I haven't actually understood the algorithm there yet. I've reached out to the authors for some clarification, but haven't gotten a response). But in general you're right. I wrote the following fuzz test:

```elm
Test.fuzz3 Fuzz.polygon2d Fuzz.int (Fuzz.floatRange 0 1) "correctly computes boundaries" <|
            \polygon edgeIndex param ->
                let
                    edges =
                        Polygon2d.edges polygon |> Array.fromList

                    edge =
                        case Array.get (edgeIndex |> modBy (Array.length edges)) edges of
                            Just e ->
                                e

                            Nothing ->
                                Debug.todo "something went wrong"

                    point =
                        LineSegment2d.interpolate edge param
                in
                polygon
                    |> Polygon2d.containsPoint point
                    |> Expect.equal Boundary
```

which fails spectacularly. So unless you can see a bug in my test code, I can remove it for a boolean.

ianmackenzie:
Ah OK - if differentiating between inside/outside/boundary is necessary for a related Boolean operation algorithm, then it might make sense to leave `containsPointHelp` and the `Containment` type as is (so they can be reused in the future, at least inside the `Polygon2d` module), stop exposing the `Containment` type, and change `containsPoint` to

```elm
contains : Point2d -> Polygon2d -> Bool
contains point polygon =
    case containsPointHelp (edges polygon) (Point2d.coordinates point) 0 of
        Inside ->
            True

        Boundary ->
            True

        Outside ->
            False
```

(In general I've been following the pattern that `contains` means 'contains point' - see `BoundingBox2d.contains`, `Triangle2d.contains`, `Sphere3d.contains` etc.) I think it probably also makes sense to take the comment referencing the paper and move it into `containsPointHelp`, since that's where the actual algorithm is.

ianmackenzie:
Also, do you know how this algorithm compares to the winding number algorithm from http://geomalgorithms.com/a03-_inclusion.html? That's the one I had tentatively settled on when doing some initial research a while ago - see #40.

gampleman:
The paper has this to say:

> The winding number algorithm [2], also called the nonzero winding number algorithm [3], includes two approaches: the first is an improved algorithm from the sum of angles that introduces the cross and dot product operations of two vectors and thereby avoids angle calculations [3]. However, if the two vectors are parallel or vertical, the calculations would fail. In addition, if the input polygon is self-intersecting or has holes, the results would be incorrect. By improving the even-odd algorithm step by step, the second approach shows the even-odd and the nonzero winding number algorithms are identical except in their interpretation of the intersection count [2].

Generally, I would assume that they both are O(n) algorithms, so not hugely different.

ianmackenzie:
Thanks! One more things I just noticed: can you change `containsPointHelp` to accept `xp` and `yp` as separate arguments instead of as a tuple? Avoiding tuple construction on each call should mean that the entire algorithm is allocation-free (in addition to already being tail recursive), which should really help performance.

tests/Tests/Polygon2d.elm
```diff
@@ -10,7 +11,8 @@ import Fuzz
 import Geometry.Expect as Expect
 import Geometry.Fuzz as Fuzz
 import LineSegment2d
-import Polygon2d
+import Point2d
+import Polygon2d exposing (convexHull)
```
ianmackenzie: `exposing (convexHull)` isn't actually necessary here, right? (And arguably confusing since `convexHull` is used as a variable name later...)

gampleman: Ah that’s probably my overzealous IDE.

ianmackenzie:
Just added a 'visual test' of the algorithm - looks good, and it seems very fast! Unless there are any other changes you wanted to make, I think this is ready to merge.

gampleman:
👍

ianmackenzie:
Merged! I'll see if I can push this out as a minor release in the next week or so (there have been a few other small additions since 1.2.1, so I'd like to do a final review and make sure the docs are all ready to go).

PR #104, opened 2019-08-08
Added Circle2d.intersectsBoundingBox and related tests.

SebastianKG:
Added a new function and tests for detecting intersections between circles and bounding boxes. Shamelessly referenced https://yal.cc/rectangle-circle-intersection-test/ for this, thanks *erkalselman* on the elmlang Slack for that recommendation.


ianmackenzie:
Thanks for the PR! On mobile so I haven't tested it yet, but a free initial comments:
- I think the `deltaX`/`deltaY` values you've calculated are actually `NearestX` and `NearestY` from the referenced web page - you're missing a `deltaX = circleX - nearestX` etc. step, which means this will only work properly for a circle at the origin (where `circleX` and `circleY` are zero).
- It would be simpler, faster and more accurate to just get `boxMaxX` and `boxMaxY` from the bounding box directly instead of reconstructing them from the min values and width/height.
- Can you add a trailing newline to AUTHORS.txt? =) I think there was one there before...

ianmackenzie:
By the way, do you only need bounding boxes or would arbitrarily-oriented rectangles be useful as well? Should be straightforward to extend this approach to rectangles, but if you don't need that functionality then I'll probably just add it in the `coordinate-systems` branch instead.

SebastianKG:
@ianmackenzie I think I've fixed the problems you pointed out -- silly mistakes like those are what I get for attempting evening programming :)

I don't need arbitrarily-oriented rectangles personally at the moment, so adding them to `coordinate-systems` is cool with me.

ianmackenzie:
Hi @SebastianKG, apologies for the delay in getting to this - I thought that modifying the code to remove the temporary `( boxMinX, boxMaxX )` and `( boxMinY, boxMaxY )` tuples would increase performance, but I wanted to take the time to test that with an actual benchmark.

On my machine, the change in the last commit I added resulted in a 2x speedup on Chrome, although only in non-optimized builds. In an optimized build there was no difference in Chrome, and on Firefox there was no difference in either build. Still, eliminating the temporary tuples never hurt performance, so might as well!

ianmackenzie:
BTW I should be pushing out a minor release with this (and a handful of other additions) soon - just doing a review of the docs now and adding a few clarifications.

PR #113, opened 2019-12-23
Adds Tests for Point2d and Point3d

matyjas:
- Point3d
-- commutative distanceFrom func
-- centroid and centroid3 matching test
- Point2d
-- interop with Tuple and Record
- elm-format 0.8.0
- reduce columns to less than 80

Pull request checklist:

- [x]   Is the pull request from a dedicated branch in your fork instead of `master`? (Preferred, but not mandatory.)
- [ ]   Have you added yourself to the AUTHORS file? Not mandatory if you'd prefer to stay anonymous, but don't be shy!
- [x]   Is "Allow edits from maintainers" enabled? (There should be a checkbox for this when you open the pull request.)
- [x]   Is the code formatted using the same version of `elm-format` as the rest of `elm-geometry`? (If running `elm-format` doesn't change any existing code, then the answer is probably 'yes'.)
- [x]   Is code (mostly) wrapped to 80 columns? (It's OK if type annotations and occasional things like string literals are longer.)
- [x]   BONUS POINTS: Have you added tests for new functionality?
- [ ]   EXTRA BONUS POINTS: Have you added documentation for new functionality? (Don't worry too much about documentation text - I like that to have a consistent style, which means writing most of it myself - but if you can come up with some code examples to use in documentation that would be fantastic.)

Don't worry too much if you're not sure if one of the above is true, or if you
know it's not true but aren't sure how to fix it - just go ahead and open the
pull request and we'll sort it out. The intent is for this list to be helpful,
not intimidating!

**Note: some of your changes snuck into my commit, i'm not sure how that happened but i hope it will merge fine**


ianmackenzie:
Looks good, thanks @matyjas!

PR #128, opened 2020-02-22
Add intersectionWithSphere function

SuzzzWood:
Added intersectionWithSphere function
[Maths from wikipedia](https://en.wikipedia.org/wiki/Line%E2%80%93sphere_intersection)

tv2


ianmackenzie:
Looks good! (And don't worry about the Travis build failure - I checked and it's an unrelated test that's failing, one that fails occasionally and I have to look into at some point but not related to any of this work.)

I was going to make a couple minor tweaks but I'm not sure I have write access to your `axis-sphere-intersection` branch - is the "Allow edits from maintainers" box checked? The changes are basically:

  - Destructure values using (e.g.) `(Types.Vector3d cto) = ...` instead of `(Vector3d cto) = ...` - this is the pattern I've used elsewhere, primarily because it seems weird to both expose the `Axis3d` type from the `Types` module _and_ have an `Axis3d` type alias in the `Axis3d` module itself
  - Use a couple functions to simplify the code: `Vector3d.componentIn` instead of `dot` to avoid the awkward direction/vector conversion, and `Point3d.along` to simplify the final point construction

SuzzzWood:
@ianmackenzie thanks for the feedback! Yes, 'Allow edits from maintainers' is checked, feel free to edit

ianmackenzie:
Just added a fuzz test for axis/sphere intersection - everything passes, so this looks good!

One last change: could you add yourself to the AUTHORS file? =)

SuzzzWood:
Nice! Have added, thanks

PR #131, opened 2020-04-30
Introduce Cone3d shape

w0rm:
This is an initial take on the `Cone3d` shape. This is mostly copied from `Cylinder3d` with some modifications:

* Renamed start and end points to base and tip points
* Changed the axis center to be the base point
* Removed `endCap`, and renamed `startCap` to  `base`
* Removed `centeredOn` and `centerPoint`
* Removed `revesrse`
* Added `basePlane`

I wonder if the API is OK, then I can work on some tests.


src/Cone3d.elm
```diff
@@ -0,0 +1,391 @@
...
+length (Types.Cone3d cone) =
+    cone.length
+
+
+{-| Get the base point of a cone. This is the center point of the circle
+that forms the base of the cone.
+-}
+basePoint : Cone3d units coordinates -> Point3d units coordinates
+basePoint cone =
+    Point3d.along (axis cone) (Quantity.multiplyBy -0.5 (length cone))
```
ianmackenzie, line 203: I think this should be just `Axis3d.originPoint (axis cone)` now =)

src/Cone3d.elm
```diff
@@ -0,0 +1,391 @@
...
+
+
+{-| Get the tip point of a cone.
+-}
+tipPoint : Cone3d units coordinates -> Point3d units coordinates
+tipPoint cone =
+    Point3d.along (axis cone) (Quantity.multiplyBy 0.5 (length cone))
```
ianmackenzie, line 210: Similarly, I think this should now be `Point3d.along (axis cone) (length cone)`

src/Cone3d.elm
w0rm: Oops! Missed to change this. Thanks for noticing.

src/Cone3d.elm
w0rm: Thanks, fixed!

ianmackenzie:
I think this API looks good!

w0rm:
@ianmackenzie thanks for the review! I will work on the tests tomorrow. Let me know if you have any ideas, otherwise I will base them on the existing cylinder tests.

ianmackenzie:
No particular test ideas off the top of my head, but if I think of anything I'll let you know =)

w0rm:
@ianmackenzie I ported a couple tests from cylinder. Let me know if anything is missing in this PR!

src/Cone3d.elm
```diff
@@ -0,0 +1,391 @@
...
+            axis cone
+
+        axialDistance =
+            Point3d.signedDistanceAlong coneAxis point
+
+        radialDistance =
+            Point3d.distanceFromAxis coneAxis point
+
+        radiusAtPoint =
+            Quantity.multiplyBy (Quantity.ratio axialDistance (length cone)) (radius cone)
```
ianmackenzie, line 265: Wait, isn't this logic reversed? Since axial distance is zero at the base of the cone, I think this should probably be

```elm
radiusAtPoint =
    Quantity.interpolateFrom (radius cone) Quantity.zero <|
        (Quantity.ratio axialDistance (length cone))
```

w0rm: Oh, good catch! Indeed when `axialDistance` is `0`, this would produce `0` radius.

w0rm: Fixed!

ianmackenzie:
I think this looks good now! I like the tests =)

Issue #134, opened 2020-06-15
Add Vector#d.scaleTo functions?


ianmackenzie:
Allow scaling a vector to a specific length (while leaving zero vectors as is), e.g.

```elm
Vector2d.scaleTo : Quantity Float units2 -> Vector2d units1 coordinates -> Vector2d units2 coordinates
```

This would be equivalent to:

```elm
scaleTo givenLength givenVector =
    case Vector2d.direction givenVector of
        Just vectorDirection ->
            Vector2d.withLength givenLength vectorDirection

        Nothing ->
            Vector2d.zero
```

This operation seems to come up fairly often, but it also seems a bit dangerous to hide inside a helper function since it makes it easy to not consider the zero-vector case.

Swendude:
I think that just adding the scaleTo definition to the documentation might do the job! This leaves the user free to change the handling of Nothing to their specific use case!

ianmackenzie:
It occurs to me that `Vector#d.scaleTo` could actually _replace_ `Vector#d.normalize` in some future breaking release - if you really wanted the current behavior of `normalize` you could replace it with `scaleTo (Quantity.float 1)`.

g-belmonte:
What would be the implication of allowing a zero direction?
Having a zero direction would make the `Maybe` disappear, and then we would have
```
scaleTo givenLength givenVector =
    Vector2d.withLength givenLength (Vector#d.direction givenVector)
```
and
```
normalize =
    scaleTo (Quantity.float 1)
```

ianmackenzie:
I mean that's exactly how things are now - "extraction of the direction from the vector" is `Vector#d.direction`, and then "dealing with scaling or having a zero vector" is some combination of `Maybe.withDefault` and `Vector#d.withLength` (or an explicit `case` expression as shown above). But the use case of "take one vector and scale it to a given length" seems to come up often enough in practice (it's come up both in discussions with others and in my own code) that it seemed useful to have a helper function for it.

g-belmonte:
Maybe if we leave a warning at the docs, saying that scaling a zero vector is equal to a zero vector? Then I think it would be good to go :)

g-belmonte:
I mean, this seems to be the most common case, so if people want a different behavior, they could easily implement it with a variation of the code snippet that you left on the first comment

ianmackenzie:
[...]

Issue #99, opened 2019-05-25
Vector2d.direction returns invalid result for very small non-zero values


MartinSStewart:
The following prints out `Direction: Just (Direction2d (Infinity,Infinity))` when the expected behavior is to print out `Direction: Nothing`
```
_ =
    Vector2d.fromComponents ( 1.0e-200, 1.0e-200 )
        |> Vector2d.direction
        |> Debug.log "Direction"
```

ianmackenzie:
Good catch! Looks like there's some floating-point underflow going on...might need a second check within `Vector2d.direction`, with some fallback logic for extremely small components.

I think it should be possible to actually return `Direction2d (0.7071..., 0.7071...)` in this case, by doing something like pre-dividing one of the two components by the other instead of doing a naive normalization.

ianmackenzie:
Basically implement this: https://d3cw3dd2w32x2b.cloudfront.net/wp-content/uploads/2012/07/Vector-length-and-normalization-difficulties.pdf

MartinSStewart:
I looked at the paper. If I'm not mistaken, it seems to introduce substantial errors into the length result? For example, the vector (2,1) has length `sqrt(2*2 + 1*1)` or `2.236...`. The paper says to divide each component by the largest component and then scale it back at the end.  So we get  `2 * sqrt(2*2 / 2 + 1*1 /2)` or `3.162...`.

ianmackenzie:
I think what the paper is suggesting is (taking your example):

- divide both components by the largest component (2), giving the vector (1, 0.5)
- find the length of that scaled vector: sqrt(1 * 1 + 0.5 * 0.5) = 1.11803
- to find the direction of the original vector, divide the scaled (1, 0.5) vector by its length (1.11803) to get (0.89443, 0.44721)
- to find the length of the original vector, multiple the length of the scaled vector (1.11803) by the largest component of the original vector (2) to get 2.23607

This works, and avoids overflow and underflow when squaring the components. I've implemented this logic in e90e37a56e1a310653461b03ce88f3248d4a8de1 in the `coordinate-systems` branch, although while doing that I realized that `Vector#d.squaredLength` was also unsafe in the same kind of way (since it can return zero or infinity for finite-length vectors). I don't think saving a single square root is enough of a performance concern to really warrant a separate function, so I've removed `Vector#d.squaredLength` and related functions like `Point#d.squaredDistanceFrom` and `LineSegment#d.squaredLength`.

[...]

PR #137, opened 2020-10-10
scaleTo methods for Vector2d and Vector3d

g-belmonte:
Pull request related to https://github.com/ianmackenzie/elm-geometry/issues/134

Pull request checklist:

    `master`?
  - [x] Have you added yourself to the AUTHORS file?
  - [x] Is the code formatted using the same version of `elm-format` as the rest of
    `elm-geometry`?
  - [x] BONUS POINTS: Have you added tests for new functionality?
  - [x] EXTRA BONUS POINTS: Have you added documentation for new functionality?

@ianmackenzie let me know what you think! :)


ianmackenzie:
Looking good, thanks for adding docs/tests! A few requests:

  - Can you try implementing `scaleTo` similar to the old `normalize`? For performance reasons I'd like to avoid the intermediate `Maybe` (extra memory allocation) if possible. I think this will involve adding something like `scaleFactor = givenLength / scaledLength` to the existing `normalize` logic and then use `x = scaleFactor * scaledX` instead of `x = scaledX / scaledLength`.
  - I think those tests would fail if the fuzzer produced a zero vector (which you can probably test by increasing the number of fuzz tests that get run, e.g. `elm-test --fuzz 5000`). I think there should probably be some logic in both the `scaleTo` and `normalize` tests that first checks if the fuzzed `vector` is equal to `Vector2d.zero`/`Vector3d.zero` and if so checks that the scaled vector is also equal to the zero vector; otherwise use the existing logic.
  - Very minor point but I think the tests would be slightly cleaner if you used `Test.fuzz2` instead of `Test.fuzz` with `Fuzz.tuple`.

g-belmonte:
I did some changes to the `scaleTo`, then I ran `elm-test --fuzz 5000`. This is what I got:

image

Is it something to worry about?

ianmackenzie:
Yeah, don't worry about that test failure, I think I just set it up in a way that you occasionally have really numerically bad situations (like an axis that's _almost_ parallel to the plane) where it's impossible to get super high accuracy no matter what you do. I do have to fix that test to avoid those cases (while being careful not to filter out _actual_ problems), but it's certainly not related to anything you're doing!

ianmackenzie:
By the way, any chance you could tweak the `Vector3d` example to be a bit more like the `Vector2d` one? (Nicer numbers, output is formatted as an expression you could actually write instead of what the REPL would show, etc.) You can 'cheat' a bit here and basically use the same 2D example by just (for example) setting the Y component to zero in the 3D case.

ianmackenzie:
For the tests, it's certainly good to have some explicit tests for the zero vector case, but I think there still needs to be something in the general case to handle the situation where the fuzzer produces a zero vector (I may not have explained this well). For example, taking `normalize` as an example because it's a bit simpler,

```elm
\vector ->
    Vector3d.normalize vector
        |> Vector3d.length
        |> Expect.quantity (Quantity.float 1)
```

will fail if `vector` happens to be `Vector3d.zero` since then the length will be zero, not one! I think you'll need something like

```elm

\vector ->
    if vector == Vector3d.zero then
        Vector3d.normalize vector
            |> Expect.equal Vector3d.zero
    
    else
        Vector3d.normalize vector
            |> Vector3d.length
            |> Expect.quantity (Quantity.float 1)
```

ianmackenzie:
Oh, and I appreciate the attempt to simplify/clean up the old implementation of `normalize`, but there's actually a reason for the weird logic with `largestComponent` 🙂. _Mathematically_ your version is the same (and simpler), but _numerically_ calculations like `x * x` can underflow or overflow with very small or very large `Float` values (see https://github.com/ianmackenzie/elm-geometry/issues/99 for a detailed discussion). Dividing everything by the largest component (and then effectively multiplying by that again afterwards, after doing the square/square root) avoids both issues.

g-belmonte:
I'm ashamed by the looks of that bad 3d vector example :joy: For the Vector2d I used Pythagorean triplets, so that I could have nice looking numbers... but I had no idea of an example like that on a 3d vector :sweat_smile: I'll put one of the components as zero for now

Now I see the problems with the tests and the implementation I did :smile: Let me fix that :)

ianmackenzie:
I think this is looking great now! Was there anything else you were thinking of adding or are you OK to have this merged?

g-belmonte:
I think this is it, at least for this one :grin:

Issue #127, opened 2020-01-25
Add velocity/acceleration/force vector constructors?


ianmackenzie:
`Vector3d.kilometersPerHour`, `Vector3d.metersPerSecondSquared`, `Vector3d.newtons` etc.

Would be quite a few extra functions but could make working with things like `elm-physics` more convenient.

ianmackenzie:
Perhaps put these in a 'Physics' section or similar towards the bottom of the documentation, so that they don't clutter things up too much?

g-belmonte:
Hmm... this seems interesting! Can I help out somehow? :smile:

ianmackenzie:
Certainly! Feel free to take a crack at it, I'll assign this issue to you =)

Don't be shy about opening up a work-in-progress pull request once you have a bit of code in place - happy to provide feedback on an initial version, suggest additional constructors that might be useful, etc.

ianmackenzie:
Implemented in #139, thanks @g-belmonte! It's possible we'll think of more units to add here in the future but I think for now this can be closed.

PR #139, opened 2020-10-10
Feature/vector physics

g-belmonte:
Pull request related to https://github.com/ianmackenzie/elm-geometry/issues/127

Pull request checklist:

    `master`?
  - [ ] Have you added yourself to the AUTHORS file? Not mandatory if you'd prefer
    to stay anonymous, but don't be shy!
  - [x] Is the code formatted using the same version of `elm-format` as the rest of
    `elm-geometry`?
  - [ ] BONUS POINTS: Have you added tests for new functionality?
  - [ ] EXTRA BONUS POINTS: Have you added documentation for new functionality?

There are still no tests and documentation. I wanted to check first if I'm going in the right direction :)

### DIFF
--- a/src/Vector2d.elm
+++ b/src/Vector2d.elm
@@ -25,6 +25,7 @@ module Vector2d exposing
     , relativeTo, placeIn
     , metersPerSecond, feetPerSecond, kilometersPerHour, milesPerHour
     , metersPerSecondSquared, feetPerSecondSquared, gees
+    , newtons, kilonewtons, meganewtons, pounds, kips
     , unsafe, unwrap
     )
 
@@ -159,6 +160,11 @@ For the examples, assume the following frame has been defined:
 @docs metersPerSecondSquared, feetPerSecondSquared, gees
 
 
+## Force
+
+@docs newtons, kilonewtons, meganewtons, pounds, kips
+
+
 # Advanced
 
 These functions are unsafe because they require you to track units manually. In
@@ -172,6 +178,7 @@ useful when writing generic/library code.
 import Acceleration exposing (MetersPerSecondSquared)
 import Angle exposing (Angle)
 import Float.Extra as Float
+import Force exposing (Newtons)
 import Geometry.Types as Types exposing (Axis2d, Direction2d, Frame2d, Point2d)
 import Length exposing (Meters)
 import Pixels exposing (Pixels)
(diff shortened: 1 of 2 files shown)


ianmackenzie:
I think for consistency with `Vector3d.meters` etc., the functions should probably have signatures like

```elm
Vector3d.metersPerSecond : Float -> Float -> Float -> Vector3d MetersPerSecond coordinates
```

I agree that the length-and-direction method is also useful, but being able to write

```elm
Vector3d.metersPerSecond 5 Direction3d.z
```

instead of

```elm
Vector3d.withLength (Speed.metersPerSecond 5) Direction3d.z
```

is not quite as much of a savings as being able to write

```elm
Vector3d.metersPerSecond 3 0 4
```

instead of what you'd currently have to do to get that:

```elm
Vector3d.xyz
    (Speed.metersPerSecond 3)
    (Speed.metersPerSecond 0)
    (Speed.metersPerSecond 4)
```

The three-argument form is also useful for JSON decoding, e.g.

```elm
Decode.map3 Vector3d.metersPerSecond
    (Decode.field "x" Decode.float)
    (Decode.field "y" Decode.float)
    (Decode.field "z" Decode.float)
```

ianmackenzie:
I'm also not sure if it makes sense to talk about a pressure vector...I think pressure is pretty fundamentally a scalar quantity. Was there a particular application you were considering?

I think _area_ vectors will come up in practice, but also I think those will usually end up being computed with code like

```elm
triangleAreaVector =
    Triangle3d.normalDirection triangle
        |> Maybe.map (Vector3d.withLength (Triangle3d.area triangle))
        |> Maybe.withDefault Vector3d.zero
```

instead of being constructed explicitly from their coordinates, so I don't think it would make sense to add `Vector3d.squareMeters` or anything like that.

g-belmonte:
Now I gotcha! I'll do some changes :grin: 

As for the pressure, yes, you're right. Pressure is always applied perpendicular to the surface. I'll remove those :)

Am I missing any other vector unit?

g-belmonte:
Changes done :)

About those area vectors, `Vector3d.squareMeters` sounds a bit weird to me, since area is an scalar :sweat_smile:
But using the normal of a surface is pretty common.
Wdyt, should I implement it too?

ianmackenzie:
I'd leave out area for now. Area vectors do come up and are useful mathematically - for example, if you multiply the area vector of a triangle (as computed above) by a scalar pressure value, then the product gives you a force value which is equal to the force vector exerted on that triangle by the given pressure! But again, I don't really see a use case for directly constructing area vectors from their XYZ components, so I would say leave them out.

I think the code is looking great now! I can't think of any other units to add (I did a general review of the units type in `elm-units` and nothing jumped out at me - and I think speed, acceleration and force will certainly be the main ones when working with `elm-physics`).

I'm happy to write docs if you'd like, but if you do want to take a crack then I think the best approach would be to have a single documentation paragraph at the top of the 'Physics' section in each of the two modules that describes the general idea and has one or two examples - I certainly don't think you need a separate description/example for each function, they're pretty self-explanatory.

g-belmonte:
I'm happy to be of assistance! :grinning: 

I was actually going to ask you for help with the docs. I've been playing around with elm-geometry and elm-units for a pair of weeks, so I can't really think of any meaningful examples or describing a good use case at the moment :sweat_smile: 
Could you take care of the docs for me then?

Also, wdyt about testing? Do you prefer to have everything tested or just a few key things? If the latter, then do you see any valuable tests I could write?

ianmackenzie:
OK just added some simple docs - see what you think!

Honestly not sure what tests would be really meaningful for these - I took a look through to check for any typos and it all looked good, so if you're happy with this then I think it's ready to merge.

g-belmonte:
The docs looks good :) 
I also couldn't think of any meaningful test, so I think it is good to go!

Issue #135, opened 2020-07-07
Add Ellipsoid3d module


ianmackenzie:
3D version of `Ellipse2d`; would be useful for 3D rendering. Should mostly be able to follow the pattern established by `Ellipse2d` and `Sphere3d`.

ianmackenzie:
See https://en.wikipedia.org/wiki/Ellipsoid for volume/surface area calculations.

g-belmonte:
I came up with this "API" while organizing my thoughts.
@ianmackenzie Let me know your thoughts :smile: 

### Constructors
```
throughPoints :
    Point3d units coordinates
    -> Point3d units coordinates
    -> Point3d units coordinates
    -> Point3d units coordinates
    -> Maybe (Ellipsoid3d units coordinates)
```

### Properties
```
centerPoint : Ellipsoid3d units coordinates -> Point3d units coordinates
```

(something to get `a`, `b` and `c` - the points where the ellipsoid intersect the axis of its own coordinate system)

```
volume : Ellipsoid3d units coordinates -> Quantity Float (Cubed units)
```

```
surfaceArea : Ellipsoid3d units coordinates -> Quantity Float (Squared units)
```

```
boundingBox : Ellipsoid3d units coordinates -> BoundingBox3d units coordinates
```

### Queries
```
contains : Point3d units coordinates -> Sphere3d units coordinates -> Bool
```

### Transformations
```
scaleAbout :
    Point3d units coordinates
    -> Float
    -> Ellipsoid3d units coordinates
    -> Ellipsoid3d units coordinates
```

```
rotateAround :
    Axis3d units coordinates
    -> Angle
    -> Ellipsoid3d units coordinates
    -> Ellipsoid3d units coordinates
```

```
translateBy :
    Vector3d units coordinates
    -> Ellipsoid3d units coordinates
    -> Ellipsoid3d units coordinates
```

```
translateIn :
    Direction3d coordinates
    -> Quantity Float units
    -> Ellipsoid3d units coordinates
    -> Ellipsoid3d units coordinates
```

```
mirrorAcross :
    Plane3d units coordinates
    -> Ellipsoid3d units coordinates
    -> Ellipsoid3d units coordinates
```

(will an ellipsoid projection always be an ellipsis? - that considering that a circle is a special case of an ellipsis)
```
projectOnto :
    Plane3d units coordinates
    -> Ellipsoid3d units coordinates
    -> Ellipsis2d units coordinates
```

```
[...]

PR #142, opened 2020-10-18
Ellipsoid3d

g-belmonte:
Pull request related to https://github.com/ianmackenzie/elm-geometry/issues/135

Pull request checklist:

master?
- [x] Have you added yourself to the AUTHORS file?
- [x] Is the code formatted using the same version of elm-format as the rest of
elm-geometry?
- [x] BONUS POINTS: Have you added tests for new functionality?
- [x] EXTRA BONUS POINTS: Have you added documentation for new functionality?

# Proposed API:

- [x] Ellipsoid3d type

### Constructors
- [x] with

### Properties
- [x] centerPoint
- [x] axes
- [x] xAxis
- [x] yAxis
- [x] zAxis
- [x] xDirection
- [x] yDirection
- [x] zDirection
- [x] xRadius
- [x] yRadius
- [x] zRadius
- [x] volume
- [ ] ~surfaceArea~
- [x] boundingBox

### Queries
- [x] contains

### Measurement
- [x] signedDistanceAlong

### Transformations
- [x] scaleAbout
- [x] rotateAround
- [x] translateBy
- [x] translateIn
- [x] mirrorAcross

### Unit conversions
- [x] at
- [x] at_

### Coordinate conversions
- [x] relativeTo
- [x] placeIn


ianmackenzie:
Just did a brief skim but looking pretty good so far!

g-belmonte:
@ianmackenzie take a look at the formula for the surface area: [wikipedia](https://en.wikipedia.org/wiki/Ellipsoid#Surface_area) or [wolfram](https://www.wolframalpha.com/input/?i=ellipsoid+surface+area)
The analytical solution seems to require elliptic integrals. Should I try to implement it like this, or should I go for the [approximate solutions](https://en.wikipedia.org/wiki/Ellipsoid#Approximate_formula)?

ianmackenzie:
Oof didn't realize that, I would just leave it out then =) I don't think it's a particularly important operation, just for spheres it's so simple it seemed silly not to include it.

g-belmonte:
@ianmackenzie the code is mostly done. We're missing top-level docs and tests.
About the tests, is there anything special you want tested? I was thinking on covering basic usage of all public functions. Wdyt?

ianmackenzie:
Hey @g-belmonte, sorry for the delay! This one took a bit of thought 🙂 Here are some cool and useful tests I can think of:

  - If you apply a random `translateBy`, `rotateAround`, `scaleAbout`, `mirrorAcross`, `placeIn` or `relativeTo` to an `Ellipsoid3d`, and the matching transformation to a `Point3d`, do you get the same result from `Ellipsoid3d.contains` before and after? For example if you have a random ellipsoid and a random point and you both move them the same amount, then if the point is contained in the ellipsoid before the transformation it should also be contained in the ellipsoid after both are transformed, and vice versa.
  - Same for `signedDistanceAlong` - if you apply a particular transformation to an ellipsoid and the same transformation to an `Axis3d` then you should get (within roundoff error) the same signed distance range before and after.
  - If you scale an ellipsoid about an arbitrary point by a scale factor `k`, does the volume increase by a factor of `k^3`?

For the first couple of tests, you can check out modules like `Tests.Generic.Curve3d` and tests like `Tests.CubicSpline3d` which use it - there are a bunch of similar tests like "if I evaluate a curve at t=0.5 and then rotate the point, it should be the same as rotating the curve and then evaluating the rotated curve at t=0.5" (which tests both curve evaluation _and_ rotation).

Let me know if that makes sense!

g-belmonte:
Yes, it does make sense! I'll work on it :grin:

ianmackenzie:
One more thought, you could probably make a couple tests based on an `Ellipsoid3d` with 3 equal radii being equivalent to a `Sphere3d` with that radius...volume should be the same, point containments results should be the same, etc. But only if you have time and it seems interesting, I don't think those are mandatory 🙂

g-belmonte:
@ianmackenzie this was the very [first test](https://github.com/ianmackenzie/elm-geometry/pull/142/files#diff-41962a3e14477f2e8964322c8ca2267f8d3506155715406491946eb4379efac5R33) I added :grin: I also wanted to compare an `Ellipsoid3d` having 2 equal radii with an `Ellipse3d`, but the latter doesn't have a `volume` function :sweat_smile: 

I had little time the last days, so I'll try to finish the tests today :)

g-belmonte:
I'm getting this message while trying to implement `Fuzz.ellipsoid3d`:
image

Should I create a new `Fuzz.frame3d` without local coordinates, or should I tweak the [type of](https://github.com/ianmackenzie/elm-geometry/pull/142/files#diff-fba5c75035b0b63d09f162da1627423322c1249bb1b81afec3c12d4682a42c55R184) `Ellipsoid3d`?

g-belmonte:
Docs and tests added. The Fuzzer thing I solved by adding the `globalFrame3d` to the `Geometry.Fuzz` module, but I haven't exposed it.

ianmackenzie:
Haven't looked through everything in detail but I think a better solution for the frame type issue would be to have `Ellipsoid3d.with` accept an arbitrary `Frame3d units coordinates defines` argument instead of a `Frame3d units coordinates {}`, and then use `Frame3d.copy` inside the constructor to make the types work out (see the implementation of `Rectangle2d.centeredOn` as an example). Then I think you should be able to drop the `globalFrame3d` fuzzer and just use `frame3d`.

g-belmonte:
@ianmackenzie done! :smile: 

I think that the PR is now ready for a thorough review. Let me know if anything!

tests/Tests/Ellipsoid3d.elm
```diff
@@ -0,0 +1,164 @@
...
+
+
+
+-- Helper methods
+
+
+ellipsoidAtOriginWithRadii : ( Float, Float, Float ) -> Ellipsoid3d Meters coordinates
+ellipsoidAtOriginWithRadii ( x, y, z ) =
+    Ellipsoid3d.with
+        { axes = Frame3d.atPoint (Point3d.meters 0 0 0)
```
ianmackenzie, line 28: I think this can just be `Frame3d.atOrigin` 🙂

tests/Tests/Ellipsoid3d.elm
```diff
@@ -0,0 +1,164 @@
...
+
+
+equivalences : Test
+equivalences =
+    Test.describe "Equivalent shapes"
+        [ Test.fuzz Fuzz.float "Ellipsoid with 3 equal radii is equivalent to a sphere" <|
+            \radius ->
+                let
+                    sphere =
+                        Sphere3d.withRadius (meters radius) (Point3d.meters 0 0 0)
```
ianmackenzie, line 46: Similarly, this can be `Point3d.origin`

ianmackenzie: (or even simpler, use `Sphere3d.atOrigin (meters radius)`...)

tests/Tests/Ellipsoid3d.elm
```diff
@@ -0,0 +1,164 @@
...
+                let
+                    beforeTransformation =
+                        Ellipsoid3d.signedDistanceAlong axis ellipsoid
+
+                    afterTransformation =
+                        Ellipsoid3d.signedDistanceAlong
+                            (Axis3d.translateBy vector axis)
+                            (Ellipsoid3d.translateBy vector ellipsoid)
+                in
+                Expect.equal True <|
```
ianmackenzie, line 121: I think you should be able to replace this with the slightly simpler
```elm
Interval.width afterTransformation |> Expect.quantity (Interval.width beforeTransformation)
```

ianmackenzie: Or to get fancier and actually check the values of both endpoints, perhaps something like
```elm
afterTransformation
    |> Expect.all
        [ Interval.minValue >> Expect.quantity (Interval.minValue beforeTransformation)
        , Interval.maxValue >> Expect.quantity (Interval.maxValue beforeTransformation)
        ]
```

src/Ellipsoid3d.elm
```diff
@@ -0,0 +1,444 @@
...
+
+        y =
+            Point3d.yCoordinateIn localFrame point
+
+        z =
+            Point3d.zCoordinateIn localFrame point
+
+        c1 =
+            Quantity.squared x
+                |> Quantity.ratio (Quantity.squared (xRadius ellipsoid))
```
ianmackenzie, line 252: I think this ratio looks backwards...`Quantity.ratio a b` is `a / b`, but using it in a pipeline like this means that you're doing `r_x^2 / x^2` instead of `x^2 / r_x^2`.

tests/Tests/Ellipsoid3d.elm
g-belmonte: Fixed! :smile: https://github.com/ianmackenzie/elm-geometry/pull/142/commits/226969520387c6b7e2e9c6e96b1004f5de2d3619

tests/Tests/Ellipsoid3d.elm
g-belmonte: Fixed! :smile: https://github.com/ianmackenzie/elm-geometry/pull/142/commits/226969520387c6b7e2e9c6e96b1004f5de2d3619

tests/Tests/Ellipsoid3d.elm
g-belmonte: Fixed! :smile: https://github.com/ianmackenzie/elm-geometry/pull/142/commits/226969520387c6b7e2e9c6e96b1004f5de2d3619

src/Ellipsoid3d.elm
g-belmonte: wow! you're right, I mixed up the ratio! Now it's fixed :smile: https://github.com/ianmackenzie/elm-geometry/pull/142/commits/178eb76bb5ead493b9693e25e8090bb9b931a072

ianmackenzie:
I think this looks good now - thanks for all your work on this!

ianmackenzie:
Anything else you were planning to do before merging? I was about to just merge but then I noticed the title was still set to 'WIP' =)

g-belmonte:
@ianmackenzie Oh, I forgot about the WIP! You can proceed with merging :)

PR #146, opened 2021-01-02
Implement Axis3d.intersectionWithRectangle

w0rm:
Implemented intersection of a ray with rectangle. I had to inline plane construction because of circular dependency.

Any idea on how this can be tested? I can add hardcoded test cases, but I wonder if we could come up with some fuzz tests.


w0rm:
Rebased and added tests similar to `Axis3d.intersectionWithTriangle`

PR #147, opened 2021-03-10
Axis triangle intersection

odf:
I think this should do it. Might need to have a closer look at how the fuzzers are implemented to make sure they hit all the interesting cases.


odf:
I think this should do it. Might need to have a closer look at how the fuzzers are implemented to make sure they hit all the interesting cases.

src/Axis3d.elm
```diff
@@ -250,6 +250,60 @@ intersectionWithPlane plane axis =
         Just (axisOrigin |> Point3d.translateIn axisDirection axialDistance)
 
 
+{-| Try to find the unique intersection point of an axis with a triangle. If
+the axis does not intersect the triangle, or if it is coplanar with it (lying
+perfectly in the plane of the triangle), returns `Nothing`.
+-}
+intersectionWithTriangle : Triangle3d units coordinates -> Axis3d units coordinates -> Maybe (Point3d units coordinates)
+intersectionWithTriangle (Types.Triangle3d ( p0, p1, p2 )) axis =
```
ianmackenzie, line 258: Minor point but I try to avoid destructuring directly in argument lists, at least in published library code - it can mess with IDEs since often they'll show the parameter names during autocomplete or when hovering over the name of a function. So my preference would be to call this just `triangle` and then destructure inside the `let`.

tests/Tests/Axis3d.elm
```diff
@@ -133,6 +136,65 @@ intersectionWithPlane =
...
+
+                direction =
+                    Vector3d.withLength (Length.meters 1) axisDirection
+
+                axisOrigin =
+                    planeIntersection
+                        |> Point3d.translateBy (Vector3d.scaleBy -t direction)
+
+                maybeAxis =
+                    Axis3d.throughPoints axisOrigin planeIntersection
```
ianmackenzie, line 173: It looks to me like you could avoid the `Maybe` here by using `Axis3d.through axisOrigin direction` (or equivalently `Axis3d.withDirection direction axisOrigin`)...

tests/Tests/Axis3d.elm
```diff
@@ -133,6 +136,65 @@ intersectionWithPlane =
...
+
+                            else
+                                Expect.fail "Expected an intersection"
+
+                        Just point ->
+                            if u < 0 || v < 0 || u + v > 1 then
+                                Expect.fail "Expected no intersection"
+
+                            else
+                                Point3d.distanceFrom point planeIntersection
```
ianmackenzie, line 193: Should be able to use `point |> Expect.point3d planeIntersection` here if you `import Geometry.Expect as Expect`

tests/Tests/Axis3d.elm
```diff
@@ -133,6 +136,65 @@ intersectionWithPlane =
...
+                ( p0, p1, p2 ) =
+                    Triangle3d.vertices triangle
+
+                e1 =
+                    Vector3d.from p0 p1
+
+                e2 =
+                    Vector3d.from p0 p2
+
+                ( u, v, t ) =
```
ianmackenzie, line 157: I think I'd be inclined to use something like 

```elm
Fuzz.tuple3
    ( Fuzz.floatRange -0.5 1.5
    , Fuzz.floatRange -0.5 1.5
    , Fuzz.floatRange -0.5 1.5
    )
```

to generate a `( Float, Float, Float )` instead of using `Fuzz.vector3d` - `Fuzz.vector3d` will generate vectors with components in the -10..10 range so _most_ of the values you generate will be out of range, very few tests will actually trigger an intersection
```

**Comment by ianmackenzie** [comment]
Left a few minor comments in the code, but I think this looks great! I like the testing strategy, that's quite clever, and I appreciate the fact that you deliberately offset the axis origin some random distance away from the triangle =)

File: src/Axis3d.elm (hunk 1 above)
Comment (odf) [reply src/Axis3d.elm]: Good point! Fixed!

File: tests/Tests/Axis3d.elm (hunk 2 above)
Comment (odf) [reply tests/Tests/Axis3d.elm]: Done...

File: tests/Tests/Axis3d.elm (hunk 3 above)
Comment (odf) [reply tests/Tests/Axis3d.elm]: and done.

File: tests/Tests/Axis3d.elm (hunk 4 above)
Comment (odf) [reply tests/Tests/Axis3d.elm:157]: Yes, great idea! This now also produces some degenerate examples that I will need to handle in the test code. Which is a good thing, obviously.

**Comment by odf** [comment]
Thanks for the great feedback. Shouldn't be too long hopefully until I've got the new failed tests fixed. :-)

**Comment by ianmackenzie** [comment]
Sounds good! Looking forward to having this in, but no rush =)

**Comment by odf** [comment]
Okay, that was awkward. I couldn't help myself but introduce a "harmless little optimization" when adjusting the code for elm-geometry. That took me far longer to figure out than I care to admit, in part because shrinking worked against me in this case rather than helping and kept leading me onto the wrong path.

Anyway, everything seems to be working nicely now, and I am perhaps infinitesimally wiser.

**Comment by odf** [comment]
It should perhaps be mentioned that the Möller–Trumbore algorithm is really just a coordinate transform that can easily be implemented using `Mat4.makeBasis`, `Mat4.inverse` and `Mat4.transform` , followed by a couple of simple range tests.

But because elm-geometry does not have a general coordinate system type right now, there's probably nothing wrong with applying Cramer's rule by hand like in the original paper.

**Comment by ianmackenzie** [comment]
That does sound like a tricky bug to track down! I do find shrinking is often actively unhelpful for geometry code - could almost certainly be improved by creating some custom shrinkers for geometric data types, but I'm tempted to just forget shrinking entirely and just use random generators combined with `Shrink.noShrink`. (This has the additional advantage that you can then reuse the random generators in other contexts, not just for testing.)

And yes, there are a couple places in `elm-geometry` which would be shorter using 4x4 matrix operations/more general coordinate system transformations, but figuring out how to support stuff like that in a type-safe way is tricky (e.g. if you apply a general matrix transformation to a `Circle2d`, the result is no longer guaranteed to be a circle). So far it hasn't come up _so_ often that I've felt the need to overhaul anything.

**Comment by ianmackenzie** [comment]
I just realized that I think it should be possible to avoid the `Quantity` unwrapping, by using `Quantity.ratio` instead of `divideBy`:

```elm
scale = 
    Vector3d.dot e1 pvec

u =
    Quantity.ratio (Vector3d.dot tvec pvec) scale

v =
    Quantity.ratio (Vector3d.dot axisDirection qvec) scale
```

I think that should work out, units-wise. (You'll also need to change `scale == 0` to `scale == Quantity.zero`.)

I realize the variety of different multiplication/division functions in `elm-units` is a bit confusing, but I think that's the price we have to pay for having such a nice simple type system 🙂

ianmackenzie:
Actually no, for that to work there would need to be a special cross product function for vectors and directions, so that the result would actually have the same units as the original vector (as opposed to `Product Unitless units`, which of course is equivalent to `units` but the compiler doesn't know that).

I guess that's something for me to think about, but in the meantime I think this looks good - OK to merge?

odf:
Actually, I can do `axisDirection = direction axis |> Vector3d.withLength (Vector3d.length e1)`, which gives it compatible units to everything else, and then the ratio operations work.

Just one more question: is there a simpler test for whether a quantity is zero then doing `Quantity.equalWithin Quantity.zero Quantity.zero scale`?

odf:
Anyway, I think this is now OK to merge, unless you are keen on making all the units match.

odf:
Ha, it turns out I can just do `Vector3d.withLength (Quantity.unsafe 1)` instead, and the compiler figures out the required units. So neat! I'm starting to like this version.

odf:
So, I decided to be a bit naughty and go with `Quantity.unwrap scale == 0` for the comparison. Let me know if you prefer something else.

I think it's ready to commit if you're happy with it.

odf:
Aaand I forgot that I don't need to unwrap anything for the comparison because `==` just does the right thing. Okay, I think I'm finally done. :-)

ianmackenzie:
Looks great, thanks! And yes, I was going to mention that you can just use `scale == Quantity.zero` in this case, but you beat me to it 🙂

PR #153, opened 2022-01-06
Poisson Disk Sampling

gampleman:
I can attempt some tests later on... Not really sure how to effectively fuzz Random generators, but it seems like there might be some reasonable properties to check for here.

Also, not sure how the docs are done, but here are some images:

This is just poisson disk sampling:

poisson disk sampling

And [here is a comparison](https://ellie-app.com/gkKkv5ZQpyva1) with  just plain random sampling:

comparison with  just plain random sampling

### DIFF
--- a/src/BoundingBox2d.elm
+++ b/src/BoundingBox2d.elm
@@ -1334,7 +1334,7 @@ poissonDiskSamplesHelp seed mode queue grid =
         PickRandomPoint ->
             let
                 n =
-                    Array.length queue
+                    Array.length queue - 1
             in
             if n == 0 then
                 Grid.values grid
--- a/src/BoundingBox2d.elm
+++ b/src/BoundingBox2d.elm
@@ -1334,7 +1334,7 @@ poissonDiskSamplesHelp seed mode queue grid =
         PickRandomPoint ->
             let
                 n =
-                    Array.length queue - 1
+                    Array.length queue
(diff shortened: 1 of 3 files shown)


gampleman:
I had a little off by one bug, that didn't affect the result, but had a rather dramatic performance impact:

Screenshot 2022-01-07 at 13 16 40

gampleman:
It seems the code is rather slow. I've implemented a few improvements (namely fixing an off-by-one error that caused some unnecessary retries and a faster array delete function which is faster by not preserving ordering).

Screenshot 2022-01-07 at 13 44 58

This has resulted in an improvement, but nothing exactly revolutionary. Similar code written in JavaScript seems much faster, but I can't see any obvious bottlenecks.

It well may be that for many use cases the code is fast enough (it certainly was for mine; I only noticed the slowness when writing some fuzz tests).

ianmackenzie:
Still TBD whether having a separate package is actually the right approach, but for now I've added `BoundingBox2d.poissonDiskSamples` and its tests to https://github.com/ianmackenzie/elm-geometry-extra, so I think this MR can (finally :sweat_smile:) be closed.

The docs are currently done by adding images to https://github.com/ianmackenzie/ianmackenzie.github.io and then linking to those - I've gone ahead and done that with the images you generated.

PR #156, opened 2023-10-14
Add dev-shell with elm 0.19.0

w0rm:
To be used in order to publish packages. The dev shell can be activated with  `nix develop .#publish`. 

The shell is created from an old nixpkgs snapshot that still has `ghc865` and [the code prior to 0.19.0 -> 0.19.1 update](https://github.com/NixOS/nixpkgs/commit/5235b190459f0fb55ebf0a05887fc89b7f1fb5de).


ianmackenzie:
Sorry for the long delay, just tested locally and everything seems to work (Elm reports the correct version, and `elm diff` runs successfully instead of reporting an error as with the normal dev shell). I do get this printed to the console when I run `nix develop .#publish`, though:

```
trace: warning: replaceChars is a deprecated alias of replaceStrings, replace usages of it with replaceStrings.
```

Do you get the same thing? Probably just something to live with, I don't see an easy way to suppress it.

w0rm:
@ianmackenzie Yeah I’m getting this too, this is because we’re importing the old nix packages that have some stuff that has been deprecated in the newer nix packages. Maybe we can patch this somehow, but it should be fine to just ignore this warning

ianmackenzie:
Yeah figured that's what it was, makes sense 🙂

